PR #6, opened 2019-06-06
readme: add design and goals information

kopiczko:


README.md
```diff
@@ -4,6 +4,16 @@
...
 This repository hosts the source code for giantswarm/architect orb.
 
+Design and goals of the project:
+
+- Replacing [architect][architect] entirely **is not** a goal. But replacing most of its functionality is.
+- Being explicit. Desire to do actual job. E.g. building a docker image and pushing it to a registry should be explicitly specified in the build.
+- Making builds understandable without looking into [architect][architect] codebase.
+- Having one line binary call on each job step - to make the outputs grouped and visible and keeping build configurations sane. Good example is [package-and-push command](https://github.com/giantswarm/architect-orb/blob/master/src/commands/package-and-push.yaml).
+- Using [architect executor][architect-executor] with the latest [architect][architect] docker image most of time.
+- Using binaries other than `architect` when appropriate (they should be available inside [architect executor][architect-executor] most of the time). E.g. `docker`, `git`, `helm`, etc. Instead of wrapping existing well known functionality in [architect][architect] binary.
```
xh3b4sd: I do not understand that executer concept. How does it work? What kind of executers could we have?

kopiczko: It's a CircleCI docker (instead of machine) executor running architect image.

TheoBrigitte: [Executors](https://circleci.com/docs/2.0/executor-types) are basically different runtime environment where a CircleCI job can run

Here are the different options:
* machine: VM running ubuntu
* macOS: VM running macOS
* docker: any docker image

Here we use a custom docker image as executor, which allow to define the set of tool we can work with from the job (architect, git, helm).

TheoBrigitte: `docker` is not available in the architect executor.

kopiczko: True, but I wanted to give direction. I it probably will be there sooner than later.

TheoBrigitte: Ok, fine with me. No blocker.

PR #7, opened 2019-06-06
readme: add info about development and releases

kopiczko:


README.md
```diff
@@ -14,6 +14,24 @@ Design and goals of the project:
...
 - Using [architect][architect] commands for complex tasks. Good example is `architect helm template` instead of awkward and error prone `sed` calls.
 
+## Development
+
+- For a completely new job that is not reusing any of the existing commands the fastest way to prototype is probably creating an inline orb. Here is an example of working inline orb prototype: https://github.com/giantswarm/release-operator/pull/121/files.
+- Create a PR and test your changes using `dev:alpha` version. This version is updated every time job `orb-tools/publish-dev` configured in [circleci config](.circleci/config.yml) runs. Which is basically on every pushed commit. Beware that if someone else is developing on another branch at the same time `dev:aplha` version is likely to be overridden by their builds. You need to coordinate that. That should be enough for time being as we don't usually develop more than one CI feature at a time. To use `dev:alpha` version you need to just change the version of the orb declaration to `architect: giantswarm/architect@dev:alpha`.
```
MarcelMue: Can it easily be switched to something like `dev:branch-name`?

README.md
```diff
@@ -14,6 +14,24 @@ Design and goals of the project:
...
+- Update [Unreleased section of CHANGELOG.md](CHANGELOG.md#Unreleased) file with the changes introduced in your PR.
+- If you want to also make a new release follow the steps in [Releases](#Releases) section (which is most likely the case) or merge your PR.
+
+## Releases
+
+1. Open a new PR with changes to `orb-tools/dev-promote-prod` job in [circleci config](.circleci/config.yml):
```
MarcelMue: All of this manual stuff could probably be done a bit more automated (env var value in circle config?) but fine for now.

kopiczko: We can do it like it's done in orb-tools-orb: https://github.com/CircleCI-Public/orb-tools-orb/blob/5f4886eaaf9c14b16ad8ee46a426c65c3103546d/.circleci/config.yml#L331

And then push tags like `master-minor-master-e232004`.

But then we wouldn't have visibility on what is released. @TheoBrigitte and I thought that what we have now is a better approach. But I agree this is far from being perfect.

How would env var approach work?

README.md
kopiczko: Looks like: https://circleci.com/docs/2.0/creating-orbs/#using-development-versions
That's a good call. I didn't know about that. I'll rewrite this point.

README.md
MarcelMue: My idea was to have an env var `RELEASE` in the build (similar to the flag we use `KEEP_RESOURCES` or how it's called) and if it's true then use `minor` and `ignore /.*/` and if false use `patch` and so on.

Not sure if it's a good improvement, but atleast it would reduce manual editing.

kopiczko: We'd lose the ability to review release that way. Otherwise I'd go with recommended way from my previous comment.

kopiczko: Something to change in the future if annoying.

PR #9, opened 2019-06-11
Fix app catalog name reference in indexing step

sslavic:


kopiczko:
Please also follow steps from releases section: https://github.com/giantswarm/architect-orb#releases

We need to create 0.1.1 release here.

sslavic:
@kopiczko 0.1.1 release has been prepared. PTAL

sslavic:
Btw, got one more confirmation that the fix is needed, EFK helm chart publishing indexed the package with broken URL, see https://github.com/giantswarm/giantswarm-incubator-test-catalog/blob/master/index.yaml

CHANGELOG.md
```diff
@@ -7,11 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.1.1] - 2019-06-14
+
+### Changed
```
kopiczko:
```suggestion
### Fixed
```

sslavic: Suggestion applied and agree with this convention.

To explain myself, https://keepachangelog.com/en/1.0.0/ examples are misleading, top ones coupled fixes under `Changed` and initially I just saw those examples; now that you suggested change I checked once more and found there are examples further below which have fixes under dedicated `Fixed` paragraph in changelog.

sslavic: @kopiczko PTAL

kopiczko: Yeah we have to get used to that.

kopiczko: Still "best" standard I have found. Feel free to suggest something more structured if you have something.

PR #14, opened 2019-07-21
Index new package only on package-and-push

sslavic:
`package-and-push` command currently re-indexes whole app catalog repo, resetting `created` timestamp for all entries when the new package is being added. This is especially problematic for test app catalogs, without semver (only sha) and the timestamp reset it's hard (for Happa) to determine and select latest package.

This PR adjusts the steps of the command to index the new package separately from the rest of the catalog repo content and then merges the two indices.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.1.2] - 2019-07-22
+
+### Fixed
+
+- Index new package only in package-and-push command.
+
 ## [0.1.1] - 2019-06-14
 
 ### Fixed
--- a/src/commands/package-and-push.yaml
+++ b/src/commands/package-and-push.yaml
@@ -17,7 +17,7 @@ steps:
   - run: |
       mkdir build && helm package ./helm/<<parameters.chart>> --destination ./build
   - run: |
-      helm repo index --url https://giantswarm.github.com/$(cat .app_catalog_name) --merge <<parameters.app_catalog>>/index.yaml ./build
+      helm repo index --url https://giantswarm.github.com/$(cat .app_catalog_name) --merge .app_catalog/index.yaml ./build
   - run: |
       cp ./build/* .app_catalog
   - run: |
--- a/src/commands/package-and-push.yaml
+++ b/src/commands/package-and-push.yaml
@@ -15,9 +15,11 @@ steps:
   - run: |
       git clone -q --depth=1 --single-branch -b master "git@github.com:giantswarm/$(cat .app_catalog_name).git" .app_catalog
   - run: |
-      helm package --save=false ./helm/<<parameters.chart>> --destination .app_catalog
+      mkdir build && helm package ./helm/<<parameters.chart>> --destination ./build
   - run: |
-      helm repo index --url https://giantswarm.github.com/$(cat .app_catalog_name) .app_catalog
+      helm repo index --url https://giantswarm.github.com/$(cat .app_catalog_name) --merge <<parameters.app_catalog>>/index.yaml ./build
+  - run: |
+      cp ./build/* .app_catalog
   - run: |
       cd .app_catalog && git add -A
   - run: |
--- a/src/commands/package-and-push.yaml
+++ b/src/commands/package-and-push.yaml
@@ -15,6 +15,8 @@ steps:
   - run: |
       git clone -q --depth=1 --single-branch -b master "git@github.com:giantswarm/$(cat .app_catalog_name).git" .app_catalog
   - run: |
+      helm init --client-only
+  - run: |
       mkdir build && helm package ./helm/<<parameters.chart>> --destination ./build
   - run: |
       helm repo index --url https://giantswarm.github.com/$(cat .app_catalog_name) --merge .app_catalog/index.yaml ./build


src/commands/package-and-push.yaml
```diff
@@ -15,9 +15,11 @@ steps:
   - run: |
       git clone -q --depth=1 --single-branch -b master "git@github.com:giantswarm/$(cat .app_catalog_name).git" .app_catalog
   - run: |
-      helm package --save=false ./helm/<<parameters.chart>> --destination .app_catalog
+      mkdir build && helm package ./helm/<<parameters.chart>> --destination ./build
   - run: |
-      helm repo index --url https://giantswarm.github.com/$(cat .app_catalog_name) .app_catalog
+      helm repo index --url https://giantswarm.github.com/$(cat .app_catalog_name) --merge <<parameters.app_catalog>>/index.yaml ./build
```
sslavic: @kopiczko can you please review and especially verify that the use of app_catalog parameter is correct on this line? I couldn't find in the docs when should one use
- just dot-syntax
- `<<parameters.param_name>>`
- `$(cat .param_name)`

sslavic:
I've tested changes by running these steps manually on my laptop with latest architect executable and with two helm versions, 2.13.2 and 2.14.2. It only worked correctly with 2.14.2; 2.13.2 it seems had a bug, `helm repo index --merge` was changing existing entries in the old index by removing quotes from `created` timestamp string.

It should work on CircleCI because architect orb packages helm 2.14.0 https://github.com/giantswarm/architect/blob/master/Dockerfile#L3
It's a bit strange that helm executable packaged in orb (2.14.0) differs from architect CLI app vendored helm package (2.13.2) but in this case it's actually good as merging wouldn't actually work. When we upgrade helm package dependency across projects it would be good to upgrade / keep in sync helm CLI in architect orb too.

src/commands/package-and-push.yaml
kopiczko: If this is path to the file I think it should be:

```suggestion
      helm repo index --url https://giantswarm.github.com/$(cat .app_catalog_name) --merge .app_catalog/index.yaml ./build
```

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.1.2] - 2019-07-22
+
+### Fixed
+
+- Index new package only in package-and-push command.
+
```
kopiczko: You can test it using `dev:index-new-package-only` version. See https://github.com/giantswarm/architect-orb#development.

Also remember to bump the version: https://github.com/giantswarm/architect-orb#releases

sslavic: @kopiczko `dev:index-new-package-only` cannot be found https://circleci.com/gh/giantswarm/prometheus-operator-app/24
I cannot find that `orb-tools/publish-dev` was triggered.

sslavic: Found it https://circleci.com/gh/giantswarm/architect-orb/219 branch name is not used for the tag - alpha tag and tag with first 8 digits of SHA are published.

sslavic: Thanks @kopiczko for tip on how to test the orb changes - uncovered that `helm package` step fails at least with these modifications because `helm init` has to be called first for helm to guarantee success of commands https://github.com/helm/helm/issues/1732#issuecomment-268093699

sslavic:
A package of prometheus operator app got pushed to incubator-test catalog [with new architect-orb](https://github.com/giantswarm/prometheus-operator-app/tree/update-architect-orb-dependency-to-0.1.2) logic.
This is the commit orb made https://github.com/giantswarm/giantswarm-incubator-test-catalog/commit/b33e0719d0c6a25c15439a2b72f062741061cdd1

Before releasing orb and merging prometheus-operator-app circleci changes to use the new orb release @pipo02mix @oponder can you please check that the timestamps / new entry in index.yaml of incbator-test catalog works now for Happa to (pre)select latest release?

pipo02mix:
Happa is taking the first element of the index, so I think we need to extend to order by created field. WDYT @oponder ?

src/commands/package-and-push.yaml
```diff
@@ -15,9 +15,13 @@ steps:
...
   - run: |
-      helm package --save=false ./helm/<<parameters.chart>> --destination .app_catalog
+      helm init --client-only
   - run: |
-      helm repo index --url https://giantswarm.github.com/$(cat .app_catalog_name) .app_catalog
+      mkdir build && helm package ./helm/<<parameters.chart>> --destination ./build
+  - run: |
+      helm repo index --url https://giantswarm.github.com/$(cat .app_catalog_name) --merge .app_catalog/index.yaml ./build
+  - run: |
+      cp ./build/* .app_catalog
```
kopiczko: Will this override index.yaml w/o any additional prompt?

sslavic: By this step ./build directory will contain merged index.yaml and chart tgz package.
Yes, copy will overwrite any existing index.yaml in .app_catalog directory. Next step commits that change of index.yaml and addition of the new package.
Both merge and copy + commit will work also on empty initialized catalog repo, even without index.yaml being there.

kopiczko:
This LGTM but we need to bump the version.

sslavic:
@kopiczko I bump version in a commit immediately after the release, and make release after merge to master so need approval for the PR. Release notes / changelog update has already been included in the PR.

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.1.2] - 2019-07-22
```
kopiczko: In response to https://github.com/giantswarm/architect-orb/pull/14#issuecomment-514150383:

> @kopiczko I bump version in a commit immediately after the release, and make release after merge to master so need approval for the PR. Release notes / changelog update has already been included in the PR. All of this follows release process docs https://github.com/giantswarm/architect-orb#releases

Then this should stay Unreleased and we should change that to 0.1.2 and add the date with the follow up PR. I don't mind releasing this within this PR but we should do it either way IMO. Yes this is nitpicky but let's build some good habits.

pipo02mix:
BTW @oponder I am providing a fix for happa

sslavic:
@pipo02mix thanks, and no rush, won't merge until confirmed it works e2e

pipo02mix:
Change in happa [implemented](https://github.com/giantswarm/happa/pull/664)

PR #15, opened 2019-07-24
Prepare 0.1.2 release

sslavic:
This PR includes one last change that was omitted in https://github.com/giantswarm/architect-orb/pull/14 as preparation for 0.1.2 release.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,5 +28,5 @@ workflows:
           requires: [orb-tools/publish-dev]
           filters:
             branches:
-              #only: master
-              ignore: /.*/
+              only: master
+              #ignore: /.*/


.circleci/config.yml
```diff
@@ -28,5 +28,5 @@ workflows:
           requires: [orb-tools/publish-dev]
           filters:
             branches:
-              #only: master
-              ignore: /.*/
+              only: master
+              #ignore: /.*/
```
kopiczko: As I wrote here https://github.com/giantswarm/architect-orb/pull/14#discussion_r306362867 this PR should change `Unreleased` to `0.1.2` in the changelog.

sslavic: https://github.com/giantswarm/architect-orb/pull/14 did that + added Unreleased on top.
This whole release process is so error prone, manual - fortunately we don't have to do it too often.

kopiczko: > #14 did that

That's what I don't like. This doesn't belong there. We can have more than one change in a release.

sslavic: Yes and this doesn't contradict. It was all in separate commits but in same PR that are merged squashed together, because we wanted to use the fix like now. Sorry but I'm not aware that there is some special release cadence for this project/repo we should adhere to, that it's not OK to release individual bug fix.

PR #17, opened 2019-07-24
push-to-docker

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/6557

Provides a new CircleCI job: `architect/push-to-docker`

Which does the following:
- get the current version we are building
- log into docker registry
- build docker image
- push docker image

This was tested here: https://circleci.com/gh/giantswarm/aws-operator/21548

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-## [0.1.3] - 2019-07-26
-
 ## Added
 
 - Add push-to-docker job.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.1.3] - 2019-07-26
+
+## Added
+
+- Add push-to-docker job.
+
 ## [0.1.2] - 2019-07-22
 
 ### Fixed
--- a/src/commands/push-to-docker.yaml
+++ b/src/commands/push-to-docker.yaml
@@ -1,14 +1,14 @@
 parameters:
-  username:
+  image:
     type: "string"
   password:
     type: "string"
   registry:
     type: "string"
-  image:
+  username:
     type: "string"
 steps:
   - run: echo -n "<<parameters.image>>:$(architect project version)" > docker_image_name
   - run: docker login --username "<<parameters.username>>" --password "<<parameters.password>>" "<<parameters.registry>>"
   - run: docker build -t "$(cat docker_image_name)" .
-  - run: docker push $(cat docker_image_name)
+  - run: docker push "$(cat docker_image_name)"
--- a/src/commands/push-to-docker.yaml
+++ b/src/commands/push-to-docker.yaml
@@ -5,8 +5,10 @@ parameters:
     type: "string"
   registry:
     type: "string"
+  image:
+    type: "string"
 steps:
-  - run: architect project version > project_version
+  - run: echo -n "<<parameters.image>>:$(architect project version)" > docker_image_name
   - run: docker login --username "<<parameters.username>>" --password "<<parameters.password>>" "<<parameters.registry>>"
-  - run: docker build -t $(cat project_version) .
-  - run: docker push $(cat project_version)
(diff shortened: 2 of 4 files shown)


src/commands/push-to-docker.yaml
```diff
@@ -0,0 +1,14 @@
...
+    type: "string"
+  registry:
+    type: "string"
+  username:
+    type: "string"
+steps:
+  - run: echo -n "<<parameters.image>>:$(architect project version)" > docker_image_name
+  - run: docker login --username "<<parameters.username>>" --password "<<parameters.password>>" "<<parameters.registry>>"
+  - run: docker build -t "$(cat docker_image_name)" .
+  - run: docker push "$(cat docker_image_name)"
```
kopiczko: I would strongly prefer to write it in this syntax:

```
  - run: |
      echo -n "<<parameters.image>>:$(architect project version)" > docker_image_name
```

Otherwise we can get into various yaml traps.

src/commands/push-to-docker.yaml
```diff
@@ -0,0 +1,14 @@
...
+    type: "string"
+  password:
+    type: "string"
+  registry:
+    type: "string"
+  username:
+    type: "string"
+steps:
+  - run: echo -n "<<parameters.image>>:$(architect project version)" > docker_image_name
+  - run: docker login --username "<<parameters.username>>" --password "<<parameters.password>>" "<<parameters.registry>>"
```
kopiczko: Using password as parameter will print it to the CI output.

kopiczko: I hope you didn't test that commit anywhere.

src/commands/push-to-docker.yaml
TheoBrigitte: When there is a need for multiline yes, but this fit perfectly in single lines here and I would prefer to keep it this way for readability.

src/commands/push-to-docker.yaml
TheoBrigitte: Password wasn't printed to CI output, also using something like `echo -n <<parameters.password>>|docker login --password-stding` is not better when it comes to prevent password to be printed IMO. Not sure how we should handle this better.

src/commands/push-to-docker.yaml
TheoBrigitte: Updated this to multilines as there are weird cases where this fails (when using pipes or so?).

src/commands/push-to-docker.yaml
TheoBrigitte: We are now using env variable names here, so value won't be exposed and we are more flexible.

TheoBrigitte:
CHANGELOG and README updated. PTALA

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.1.3] - 2019-07-26
```
kopiczko: I'd like to leave it in unreleased. Let's change that in the release PR.

src/jobs/push-to-docker.yaml
```diff
@@ -0,0 +1,28 @@
+executor: architect
+parameters:
+  image:
+    description: "Name of the docker image, without tag"
+    type: "string"
+  password_var:
```
kopiczko: How about `password_envvar`?

CHANGELOG.md
TheoBrigitte: done

src/jobs/push-to-docker.yaml
```diff
@@ -0,0 +1,28 @@
...
+    default: "ARCHITECT_DOCKER_REGISTRY_PASSWORD"
+    description: "Environment variable name holding docker registry password"
+    type: "string"
+  registry:
+    default: "quay.io"
```
kopiczko: How about this registry thing? Can we take that from the image?

src/jobs/push-to-docker.yaml
TheoBrigitte: More accurate, going with this.

src/jobs/push-to-docker.yaml
TheoBrigitte: So splitting the string from image, like this ?
```
$ echo -n 'quay.io/giantswarm/aws-operator'|cut -d/ -f1
quay.io
```

Why not, no strong opinion here.

PR #20, opened 2019-07-26
release new version 0.2.0

TheoBrigitte:
Towards: giantswarm/giantswarm#6557

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,9 +24,9 @@ workflows:
       - orb-tools/dev-promote-prod:
           orb-name: giantswarm/architect
           publish-token-variable: CIRCLECI_DEV_API_TOKEN
-          release: patch
+          release: minor
           requires: [orb-tools/publish-dev]
           filters:
             branches:
-              #only: master
-              ignore: /.*/
+              only: master
+              #ignore: /.*/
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.2.0] - 2019-07-26
```
kopiczko: I'm thinking if we shouldn't also include changes for catalogbot env variable in this release. But I won't have time to work on that today :(

TheoBrigitte: I think it is fine to do this later in a major version. Here we have a minor version with a new functionality, but refactoring and breaking change will go into a major version. Fine IMO.

.circleci/config.yml
```diff
@@ -24,9 +24,9 @@ workflows:
...
           publish-token-variable: CIRCLECI_DEV_API_TOKEN
-          release: patch
+          release: minor
           requires: [orb-tools/publish-dev]
           filters:
             branches:
-              #only: master
-              ignore: /.*/
+              only: master
+              #ignore: /.*/
```
kopiczko: Don't forget to rollback changes to this file in a follow up PR.

PR #23, opened 2019-08-29
Added retries and pull for race condition

ferrandinand:
Add retries for pushing apps to avoid race conditions and unmerged changes.

### DIFF
--- a/src/commands/package-and-push.yaml
+++ b/src/commands/package-and-push.yaml
@@ -27,4 +27,4 @@ steps:
   - run: |
       cd .app_catalog && git commit -m "add $(git status --porcelain | cut -c4- | grep <<parameters.chart>>)"
   - run: |
-      cd .app_catalog && git push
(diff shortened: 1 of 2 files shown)


src/commands/package-and-push.yaml
```diff
@@ -27,4 +27,4 @@ steps:
   - run: |
       cd .app_catalog && git commit -m "add $(git status --porcelain | cut -c4- | grep <<parameters.chart>>)"
   - run: |
-      cd .app_catalog && git push
+      for i in {1..3}; do cd .app_catalog && git push && break || git pull && sleep 5; done
```
JosephSalisbury: I might be being dumb here, but wouldn't this mean that:
- change directory to `.app_catalog`
- push fails
- git pull, sleep 5
- change directory to non-existent `app_catalog`?

ferrandinand:
We are going to change the approach to rebuild all the job

PR #28, opened 2019-09-13
commands, jobs: add go-build job

kopiczko:
- Example use (devctl): https://github.com/giantswarm/devctl/pull/24
- Exemple build (devctl): https://circleci.com/gh/giantswarm/devctl/125?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link


src/commands/go-build.yaml
```diff
@@ -0,0 +1,38 @@
...
+    type: "string"
+steps:
+  - restore_cache:
+      keys:
+        - go-build-go-mod-v1-{{ checksum "go.sum" }}
+  - run: |
+      architect project version > .project_version
+  - run: |
+      cat .project_version
+  - run: |
```
MarcelMue: Why do all of these need to be different `run` steps? Can they not be done in one step?

kopiczko: I was talking about this during orb walkthrough. This is so we have output of each command in a different block and know which of them actually failed. See example build: https://circleci.com/gh/giantswarm/devctl/125?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

MarcelMue: I get it and it makes sense, the code still makes my eyes bleed a bit. Not blocking.

kopiczko: This is ~10 bash commands. I would argue you can get trough 10 lines with a bit bleeding eyes :) IMO this is way simpler to both implement and understand than the current architect approach which is running most of those command under the hood anyway.

PR #34, opened 2019-09-20
Simplify release process

TheoBrigitte:
This PR simplifies the process to release this orb to CircleCI.

Before we had:
- 1 PR for changes to changelog + enable release
- 1 PR to revert "enable release"
- tag version

Now:
- 1 PR for changes to changelog
- tag version

Also major benefit here is that the git tag is what will be used as the orb version. So we no longer need to guess what version we are publishing, it's just the git tag.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,10 @@ orbs:
 workflows:
   catalog_build:
     jobs:
-      - orb-tools/lint
+      - orb-tools/lint:
+          filters:
+            tags:
+              only: /.*/
 
       - orb-tools/pack:
           source-dir: src
@@ -14,6 +17,9 @@ workflows:
           workspace-path: packed/orb.yml
           artifact-path: packed/orb.yml
           requires: [orb-tools/lint]
+          filters:
+            tags:
+              only: /.*/
 
       - orb-tools/publish-dev:
           orb-name: giantswarm/architect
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -6,27 +6,35 @@ orbs:
...
           requires: [orb-tools/pack]
 
-      - orb-tools/dev-promote-prod:
-          orb-name: giantswarm/architect
+      - orb-tools/publish:
+          attach-workspace: true
+          orb-path: workspace/packed/orb.yml
+          orb-ref: giantswarm/architect@${CIRCLE_TAG##v}
           publish-token-variable: CIRCLECI_DEV_API_TOKEN
-          release: patch
```
kopiczko: I have a feeling that this will always bump the patch version, no?

TheoBrigitte: We use `orb-tools/publish` and not `orb-tools/dev-promote-prod` here. The difference is that instead of specifying with part to bump, you just pass the version. So I pass the git tag in there.

kopiczko: Ah. With you. Awesome!

.circleci/config.yml
```diff
@@ -6,27 +6,35 @@ orbs:
...
-          requires: [orb-tools/publish-dev]
+          requires: [orb-tools/pack]
           filters:
             branches:
-              #only: master
               ignore: /.*/
+            tags:
+              only: /.*/
```
kopiczko: For this job I think we should be extra safe use some better filtering for versions. Like `/^v\d+\.\d+\.\d+$/`.

TheoBrigitte: Yeah good point.

Just for info circleci also validates the tag format:
```
Error: Invalid version specified for publishing: my-test (expected either a semver-style version string in x.y.z format, or a development-style version string in dev:* format.)
```

PR #35, opened 2019-09-23
Simplify dev versions

TheoBrigitte:
After simplifying the release process https://github.com/giantswarm/architect-orb/pull/34 I though it would be cool to also simplify the dev process.

So instead of using `dev:<git-sha>` we can just use `dev:<git-branch>` and remove the need to change the version when we want to test our changes.

This PR replace the `orb-tools/publish-dev` job with `orb-tools/publish` and use `dev:${CIRCLE_BRANCH}` as the orb version.

I tested this and learned that git branch names can be quite funny:
```
$ circleci orb source 'theo/hello@dev:th!s)_is]/a>_va|!d&_branch<_%name+#'
commands:
  hello:
    steps:
    - run: echo hello orb
description: |
  Hello orb.
  code: https://github.com/TheoBrigitte/hello-orb.
version: 2.1
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,13 +21,21 @@ workflows:
             tags:
               only: /.*/
 
-      - orb-tools/publish-dev:
-          orb-name: giantswarm/architect
+      - orb-tools/publish:
+          name: publish-branch
+          attach-workspace: true
           orb-path: workspace/packed/orb.yml
+          orb-ref: giantswarm/architect@dev:${CIRCLE_BRANCH}
           publish-token-variable: CIRCLECI_DEV_API_TOKEN
           requires: [orb-tools/pack]
+          filters:
+            branches:
+              only: /.*/
+            tags:
+              ignore: /.*/
 
       - orb-tools/publish:
+          name: publish-tag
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -21,13 +21,21 @@ workflows:
...
-      - orb-tools/publish-dev:
-          orb-name: giantswarm/architect
+      - orb-tools/publish:
+          name: publish-branch
+          attach-workspace: true
           orb-path: workspace/packed/orb.yml
+          orb-ref: giantswarm/architect@dev:${CIRCLE_BRANCH}
           publish-token-variable: CIRCLECI_DEV_API_TOKEN
           requires: [orb-tools/pack]
+          filters:
```
MarcelMue: I get it but it's confusing on first glance. 
So the first `orb-tools/publish` is only running on branches, the second one only on tags.

TheoBrigitte: Yes, that why I added job names to better distinguish them `publish-branch` and `publish-tag`.

kopiczko: This is the syntax and I doubt we can do much.

PR #42, opened 2019-10-03
go-build: add os parameter

kopiczko:


kopiczko:
Tested with https://github.com/giantswarm/opsctl/pull/442

PR #51, opened 2019-12-10
add-old-push-docker

calvix:
add new job for pushing the images the old way ( without version, tag is just SHA)
This is for pushing CP components images to china

### Checklist

- [ ] Update the [appcatalog docs](https://github.com/giantswarm/giantswarm/blob/master/processes/appcatalog.md) to use the new release.

towards https://github.com/giantswarm/giantswarm/issues/7435

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add "tag-latest-branch" parameter to the push-to-docker job.
+- Add "push-to-docker-legacy" command to be able push old style docker image tags.
 
 ## [0.4.3] 2019-10-10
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add "tag-latest-branch" parameter to the push-to-docker job.
+- Add "push-to-docker-legacy" command to be able push old style docker image tags.
```
rossf7: I think 0.4.4 is already released so this should be a new release.

README.md
```diff
@@ -206,6 +206,10 @@ workflows:
               only: /^v.*/
 ```
 
+### push-to-docker-legacy
+
+Same as `push-to-docker` with only change that the docker image tag is only including SHA of commit instead of version + SHA.
```
rossf7:
```suggestion
Same as `push-to-docker` with only change that the docker image tag only includes the commit SHA instead of version + SHA.
```

PR #55, opened 2020-01-31
push-to-app-collection: add "unique" parameter

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/7648

If `unique` is specified then:

- All other app files will be removed from the collection with `git rm APP-*.yaml`
- This app will be added with `git add APP-unique.yaml`

Note I decided to replace `APP-1.2.3.yaml` with `APP-unique.yaml`. This is because I'd like to have an update instead of replacement in case of unique app. When this is a problem? E.g. with OPA gatekeeper (which we make this feature for) behaviour is undefined (at least to me) when two of them are installed at the same time. So when one app CR is deleting and other creating we don't have a guarantee that there won't be two deployments at the same time.

### DIFF
--- a/src/jobs/push-to-app-collection.yaml
+++ b/src/jobs/push-to-app-collection.yaml
@@ -1,6 +1,8 @@
 description: >
   Generates and adds an App CR to an app-collection.
   Note: version will be detected automatically by architect.
+
+
 executor: architect
 parameters:
   app_name:
@@ -14,7 +16,7 @@ parameters:
     description: "Name of the app collection repository."
     type: "string"
   unique:
-    description: "Whether the previous App versions in the collection should be replaced with this one. Any non-empty string means true."
+    description: "Whether the previous App versions in the collection should be replaced with this one. Any non-empty string means true. This is useful when only one (usually latest) version of the app should be installed at a time. E.g. OPA gatekeeper."
     type: "string"
     default: ""
 steps:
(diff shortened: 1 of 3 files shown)


jgsqware:
> Note I decided to replace APP-1.2.3.yaml with APP-unique.yaml

This means we cannot see which version can be installed?

kopiczko:
> > Note I decided to replace APP-1.2.3.yaml with APP-unique.yaml
> 
> This means we cannot see which version can be installed?

You can always open the file. E.g. https://github.com/giantswarm/aws-app-collection/blob/349150be26f5882d02afa994beff09e8430a3262/helm/aws-app-collection-chart/templates/aws-operator-8.1.0.yaml#L34

src/jobs/push-to-app-collection.yaml
```diff
@@ -4,19 +4,24 @@ description: >
...
-    description: "Name of the catalog CR where the application belongs"
+    description: "Name of the catalog CR where the application belongs."
     type: "string"
     default: "control-plane-catalog"
   app_collection_repo:
-    description: "Name of the app collection repository"
+    description: "Name of the app collection repository."
     type: "string"
+  unique:
+    description: "Whether the previous App versions in the collection should be replaced with this one. Any non-empty string means true."
```
JosephSalisbury: It'd be useful to document _when_ to use this. e.g: this is useful for applications which we don't support running multiple versions of (e.g: gatekeeper)

PR #58, opened 2020-02-03
Goimports local

ubergesundheit:
Towards https://github.com/giantswarm/giantswarm/issues/7485

Adds a step to the `go-test` command which runs the command suggested by giantswarm/giantswarm/issues/7485 on *.go files except for the vendor directory.

I think this won't make the job fail but will print offending files in the log. I think some old files are not following the convention defined in fmt. (https://github.com/giantswarm/fmt/blob/master/go/imports.md)

### DIFF
--- a/src/commands/go-test.yaml
+++ b/src/commands/go-test.yaml
@@ -19,8 +19,7 @@ steps:
       echo -n " -X '<< parameters.pkg >>.version=$(cat .project_version)'" >> .ldflags
   - run: |
       echo "\"$(cat .ldflags)\""
-  - name: Check if imports are properly sorted
-    run: |
+  - run: |
       go get golang.org/x/tools/cmd/goimports && [[ -n $(goimports -local github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME} -l .) ]] && goimports -local github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME} -d . && exit 1
   - run: |
       test -z $(gofmt -l .) || gofmt -d .
--- a/src/commands/go-test.yaml
+++ b/src/commands/go-test.yaml
@@ -6,6 +6,8 @@ steps:
   - run: |
       go mod tidy && git diff --exit-code
   - run: |
+      find . -path ./vendor -prune -o -name '*.go' -exec goimports -local github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME} -d "{}" +
+  - run: |
       architect project version > .project_version
   - run: |
       cat .project_version
--- a/src/commands/go-test.yaml
+++ b/src/commands/go-test.yaml
@@ -19,8 +19,10 @@ steps:
       echo -n " -X '<< parameters.pkg >>.version=$(cat .project_version)'" >> .ldflags
   - run: |
       echo "\"$(cat .ldflags)\""
-  - run: |
-      go get golang.org/x/tools/cmd/goimports && [[ -n $(goimports -local github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME} -l .) ]] && goimports -local github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME} -d . && exit 1
+  - run:
+      name: Check if imports are properly sorted
+      command: |
+        go get golang.org/x/tools/cmd/goimports && [[ -n $(goimports -local github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME} -l .) ]] && goimports -local github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME} -d . && exit 1
   - run: |
       test -z $(gofmt -l .) || gofmt -d .
   - run: |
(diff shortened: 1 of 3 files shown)


src/commands/go-test.yaml
```diff
@@ -5,6 +5,8 @@ parameters:
 steps:
   - run: |
       go mod tidy && git diff --exit-code
+  - run: |
+      find . -path ./vendor -prune -o -name '*.go' -exec goimports -local github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME} -d "{}" +
```
kopiczko: I think we need to install `goimports` first.

kopiczko:
>  think this won't make the job fail

I think it'd be better to make the job fail.

ubergesundheit:
> Are you sure you linked correct issue? ([giantswarm/giantswarm#748](https://github.com/giantswarm/giantswarm/issues/748))

Should be giantswarm/giantswarm#7485 :)

src/commands/go-test.yaml
```diff
@@ -19,6 +19,8 @@ steps:
       echo -n " -X '<< parameters.pkg >>.version=$(cat .project_version)'" >> .ldflags
   - run: |
       echo "\"$(cat .ldflags)\""
+  - run: |
+      go get golang.org/x/tools/cmd/goimports && find . -path ./vendor -prune -o -name '*.go' -exec goimports -local github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME} -d "{}" +
```
kopiczko: Projects using this orb are using go modules. So I guess we don't have to do the find magic.

But still I'd prefer to make this somehow fail if goimports thinks imports are not fine.

xh3b4sd: Make it fail 💯

ubergesundheit: :fire:

src/commands/go-test.yaml
```diff
@@ -19,6 +19,8 @@ steps:
       echo -n " -X '<< parameters.pkg >>.version=$(cat .project_version)'" >> .ldflags
   - run: |
       echo "\"$(cat .ldflags)\""
+  - run: |
+      go get golang.org/x/tools/cmd/goimports && [[ -n $(goimports -local github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME} -d .) ]] && exit 1
```
kopiczko: I think it would be good to display the output (`-l` should be a bit cheaper to run for checks - I'm not super fussed).

```suggestion
      go get golang.org/x/tools/cmd/goimports && [[ -n $(goimports -local github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME} -l .) ]] && goimports -local github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME} -d . && exit 1
```

ubergesundheit: :+1:

src/commands/go-test.yaml
```diff
@@ -19,6 +19,8 @@ steps:
       echo -n " -X '<< parameters.pkg >>.version=$(cat .project_version)'" >> .ldflags
   - run: |
       echo "\"$(cat .ldflags)\""
+  - run: |
+      go get golang.org/x/tools/cmd/goimports && [[ -n $(goimports -local github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME} -l .) ]] && goimports -local github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME} -d . && exit 1
```
marians: Would be cool if we could start using the name/command tuple here as explained in https://github.com/giantswarm/fmt/blob/master/circleci/all-circleci-best-practices.md

ubergesundheit: Where does the schema live? CI fails because `name` is not defined in the schema. Or did I do something wrong?

```
Error: ERROR IN CONFIG FILE:
[#/commands/go-test] 0 subschemas matched instead of one
1. [#/commands/go-test] expected type: String, found: Mapping
|   Command may be a string reference to another command
2. [#/commands/go-test/steps/8] 0 subschemas matched instead of one
|   1. [#/commands/go-test/steps/8] expected type: String, found: Mapping
|   |   Shorthand commands, like `checkout`
|   |   SCHEMA:
|   |     type: string
|   |   INPUT:
|   |     name: Check if imports are properly sorted
|   |     run: |
|   |       go get golang.org/x/tools/cmd/goimports && [[ -n $(goimports -local github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME} -l .) ]] && goimports -local github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME} -d . && exit 1
|   2. [#/commands/go-test/steps/8] maximum size: [1], found: [2]
|   |   long form commands like `run:`
|   |   SCHEMA:
|   |     maxProperties: 1
|   |   INPUT:
|   |     name: Check if imports are properly sorted
|   |     run: |
|   |       go get golang.org/x/tools/cmd/goimports && [[ -n $(goimports -local github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME} -l .) ]] && goimports -local github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME} -d . && exit 1
```

ubergesundheit: It seems circleci does not allow a `name` key here. Reverting this now..

ubergesundheit:
May I ask for an approval?

PR #61, opened 2020-02-03
push-to-app-collection: fix unique CR name

kopiczko:


JosephSalisbury:
For my understanding, why is this a problem?

kopiczko:
> For my understanding, why is this a problem?

So when app cr named `gatekeeper-0.1.0`is replaced with app cr `gatekeeper-0.2.0` you don't have guarantee that the 0.1.0 app will be deleted before 0.2.0 is installed. If instead you update version from 0.1.0 to 0.2.0 in `gatekeeper-unique` app cr you have that guarantee.

kopiczko:
And I don't want to have more than one gatekeeper deployment installed at a time. I have no idea what happens in that scenario (and I don't feel the urge to discover).

PR #65, opened 2020-02-07
Add feature to test Helm charts

yasn77:
Using chart testing (https://github.com/helm/chart-testing), add job to lint
Helm charts.

Chart testing is pretty extensive and we can easily add other tests using it.

IMHO this is the first step, next would be to remove the executor by adding `ct`
to architect docker image. After that look to extend architect command itself to
either call ct or use ct as a module (not sure if that is possible).

### Checklist

- [ ] Update the [appcatalog docs](https://github.com/giantswarm/giantswarm/blob/master/areas/managed-services/adding_app_to_appcatalog.md) to use the new release.


yasn77:
Looks like test failed talking to Github API :man_shrugging:

yasn77:
I might mark this as a draft as there is a bit of an issue with it. The actual linting job relies on a different Docker image, and that causes an issue with templating first then linting.... 

I have a [branch in `kong-app`](https://circleci.com/gh/giantswarm/kong-app/tree/use_chart_testing) that's using these jobs

src/executors/helm-chart-testing.yaml
```diff
@@ -0,0 +1,3 @@
+---
+docker:
+  - image: quay.io/helmpack/chart-testing:v2.4.0
```
kopiczko: Why do we need a separate executor here?

src/commands/helm-ct-lint.yaml
```diff
@@ -0,0 +1,18 @@
+---
+description: Lint Helm charts
+parameters:
+  config_file:
+    type: string
+    default: ""
```
kopiczko: Let's not specify default in the command.

```suggestion
```

src/commands/helm-ct-lint.yaml
```diff
@@ -0,0 +1,18 @@
...
+  additional_args:
+    type: string
+    default: ""
```
kopiczko:
```suggestion
```

src/executors/helm-chart-testing.yaml
kopiczko: How about renaming this fine `chelmpack-chart-testing.yaml` to have something more structured?. We also need to retag this image.

src/jobs/helm-template.yaml
```diff
@@ -0,0 +1,9 @@
+---
```
kopiczko: Why do we need this job?

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Functionality to run Helm linting using Chart testing (https://github.com/helm/chart-testing)
```
kopiczko: I'd like to have lines saying what jobs were added. Like `Add helm-ct-lint`.

src/commands/helm-ct-lint.yaml
```diff
@@ -0,0 +1,18 @@
+---
```
kopiczko: I'd remove `-ct` part from the name. I think tooling we use here is irrelevant. That way in the future we can change the tooling and not change the name remaining backward compatible.

src/executors/helm-chart-testing.yaml
yasn77: Helm chart testing tool isn't apart of architect image. So unless the command/job includes steps to install it, we can't use the existing executor.  I felt, using this executor would be simpler than replicating what it does in a command/job. 

I was intending to raise PR against architect repo to include them but wasn't really if it would be needed.

src/commands/helm-ct-lint.yaml
yasn77: My understanding is, that by removing `default` means that the option is required (which it isn't)

src/commands/helm-ct-lint.yaml
yasn77: Same as above.

src/executors/helm-chart-testing.yaml
kopiczko: Yeah I figured later and forgot to delete this comment. I think it's ok to use the image. It should be retagged but I left another comment for that.

yasn77: I'll retag and rename

src/commands/helm-ct-lint.yaml
kopiczko: You're right but we always pass the parameter from the job. I'd rather not bother with keeping defaults in sync in multiple places.

src/jobs/helm-template.yaml
yasn77: Because of the executor problem. Steps run in the same executor :/

So had to create a different job. Although this causes a different problem which I am looking at.

kopiczko: So this is kind of sacrifice of DX for easiness of implementation? I don't like that. Would installing all the required tooling in architect docker image solve the problem?

yasn77: @kopiczko Yep, I agree it's not really 'elegant' and yeh it would be fixed by adding it to architect docker image (which I am going to do).

I am going to close this PR and reopen once changes are in the architect docker image

yasn77:
Closing and will reopen once architect docker image contains `ct` (helm chart testing)

cokiengchiara:
@yasn77 I'm going through PDV checklist - is this still valuable? Or should just move to archive?

yasn77:
> @yasn77 I'm going through PDV checklist - is this still valuable? Or should just move to archive?

@cokiengchiara I reopened this in #80

cokiengchiara:
sounds good, will just remove this from board then!

PR #69, opened 2020-02-13
fix push-to-app-collection unique parameter

TheoBrigitte:
This PR fixes an issue with the `unique` parameter in the `push-to-app-collection` job.

When a new app with the unique flag is added to the app-collection, there is no previous app with the same name to be cleaned up and the job fails with the following error:

image

see https://circleci.com/gh/giantswarm/gatekeeper-control-plane-rules/46


kopiczko:
Update changelog please <3

PR #70, opened 2020-02-17
go-test: stop injecting version into build info

mig4:
Towards https://github.com/giantswarm/giantswarm/issues/8637

We want to invert how the project version is maintained and instead of
generating it from git data and injecting during the build, we want to
maintain it in source files. This will allow managing releases by
merging PRs and that they'll undergo a review process.

### DIFF
--- a/src/commands/go-test.yaml
+++ b/src/commands/go-test.yaml
@@ -6,17 +6,11 @@ steps:
   - run: |
       go mod tidy && git diff --exit-code
   - run: |
-      architect project version > .project_version
-  - run: |
-      cat .project_version
-  - run: |
       echo -n "-w -linkmode 'auto' -extldflags '-static'" > .ldflags
   - run: |
       echo -n " -X '<< parameters.pkg >>.buildTimestamp=$(date --utc '+%FT%TZ')'" >> .ldflags
   - run: |
       echo -n " -X '<< parameters.pkg >>.gitSHA=${CIRCLE_SHA1}'" >> .ldflags
-  - run: |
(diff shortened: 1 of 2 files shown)


src/commands/go-test.yaml
```diff
@@ -15,8 +15,6 @@ steps:
       echo -n " -X '<< parameters.pkg >>.buildTimestamp=$(date --utc '+%FT%TZ')'" >> .ldflags
   - run: |
       echo -n " -X '<< parameters.pkg >>.gitSHA=${CIRCLE_SHA1}'" >> .ldflags
-  - run: |
-      echo -n " -X '<< parameters.pkg >>.version=$(cat .project_version)'" >> .ldflags
```
MarcelMue: Why still have line 9 in this file? Or was it just missed?

JosephSalisbury: And 11 too, I guess

mig4: Yeah, I missed it, the file doesn't seem to be used anywhere else so it's gone now :wink:

mig4: Thanks for catching that!

mig4:
Tested on [prometheus-meta-operator](https://github.com/giantswarm/prometheus-meta-operator/tree/test-architect-orb-dev) - [build job](https://app.circleci.com/jobs/github/giantswarm/prometheus-meta-operator/338) (compared to [this one](https://app.circleci.com/jobs/github/giantswarm/prometheus-meta-operator/332) from master branch it doesn't inject `version`).

PR #71, opened 2020-02-18
changelog-lint: add the job

kopiczko:
Tested within the PR:
- commit: https://github.com/giantswarm/architect-orb/pull/71/commits/2334b0fe21ac784907d6b148960e99ab71da1158
- build: https://circleci.com/gh/giantswarm/architect-orb/914?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Fixed in:
- commit: https://github.com/giantswarm/architect-orb/pull/71/commits/4af9000d8e8b72203e155ef1edf3bc3eb9df1ea0 
- build: https://circleci.com/gh/giantswarm/architect-orb/922?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

PR #75, opened 2020-02-20
Support dep

tfussell:
Since several people have been noticing unsorted imports which would have been caught by `goimports`, and `goimports` isn't supported for `dep`-based projects using `architect-orb`, I decided to try adding it.

It has been tested as an inline orb in https://github.com/giantswarm/apiextensions/pull/330 for testing, building, and deploying. The CircleCI build can be seen here https://circleci.com/workflow-run/23937de0-da62-4e07-b589-607b8ed10394.

While it's not ideal spending time supporting systems we're moving away from, this should speed up PR iteration loops for `dep`-based projects by catching issues we have to look for manually now (or miss entirely).

This orb branch has been tested here https://circleci.com/workflow-run/8dc9eadd-5a4d-4ee4-9d50-eb140312f64d. To ensure that there was no regression, I've also tested it with a project that uses modules, `ignition-operator`, here https://circleci.com/workflow-run/80239c71-c449-4daa-8c77-88c72f452721

## Checklist


src/jobs/go-architect.yaml
```diff
@@ -0,0 +1,12 @@
+description: |
```
kopiczko: Why do we need this?

src/commands/go-test-legacy.yaml
```diff
@@ -0,0 +1,48 @@
...
+        ln -s /go/src/project /go/src/github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}
+  - run:
+      name: Installing Dep
+      command: |
+        go get github.com/golang/dep/cmd/dep
+  - run:
+      name: Running Dep Ensure
+      command: |
+        export DEPPROJECTROOT=github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}
+        dep ensure && git diff --exit-code
```
kopiczko: I'd skip that part. I recon this will take ridiculous amount of time.

src/commands/go-test-legacy.yaml
```diff
@@ -0,0 +1,48 @@
+steps:
+  - run:
+      name: Configuring Go Environment
```
kopiczko: Can we follow this pattern for names: https://github.com/giantswarm/architect-orb/pull/71/files#diff-9f10d1e31cf5e578c665e884ed6d3d02R3

I think this is more useful for the orb. I want to document that but I haven't found time yet.

src/commands/go-test-legacy.yaml
```diff
@@ -0,0 +1,48 @@
...
+  - run:
+      name: Running Go Vet
+      command: |
+        cd /go/src/github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}
+        CGO_ENABLED=0 go vet ./...
+  - run:
+      name: Running Go Test
+      command: |
+        cd /go/src/github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}
+        CGO_ENABLED=0 go test ./...
```
kopiczko: I think the testing part is rather simple. Do you have plans for the `go-build-legacy` job? How about pushing charts to quay?

src/jobs/go-architect.yaml
tfussell: A description? It's a sort of documentation for people setting up new workflows. CircleCI orbs are public so it makes it look nicer here https://circleci.com/orbs/registry/?filterBy=all&query=giantswarm. And other jobs already have it so I was trying to be consistent.

kopiczko: I meant this job :D

kopiczko: Sorry that was quite a shortcut.

src/commands/go-test-legacy.yaml
tfussell: CircleCI will terminate the job if it hits 10 minutes without output so at least it's not unbounded. I think there's value in ensuring that someone didn't forget to update their dependencies and vendor directory after modifying `Gopkg.toml` and that `vendor/` hasn't been manually edited.

src/commands/go-test-legacy.yaml
tfussell: I was trying to follow the CircleCI style. They have job names like `Spin Up Environment` and `Persisting to Workspace`. Prefixing with the orb command name seems reasonable though.

src/commands/go-test-legacy.yaml
kopiczko: Ah this is ensure w/o -update. Yes this should be fine.

src/commands/go-test-legacy.yaml
kopiczko: Yeah I know what you mean. We also have fmt entry about that. But I'd rather be practical here.

src/jobs/go-architect.yaml
tfussell: Heh, okay. This approach means we don't need to download the architect binary on the CircleCI runner. We can run it in a container and get a known version of architect. It also means less to get out of sync between repos which currently need to implement this step individually in each CircleCI config. We can go ahead and switch all repos to use this orb and then migrating to go modules is easier.

src/commands/go-test-legacy.yaml
tfussell: I wrote this `go-build-legacy` job at first, and then decided to wrap it up into a `go-architect` job instead. There's a lot of logic in `./architect build` workflows and I didn't want to change too much in the orb.

src/jobs/go-architect.yaml
kopiczko: I understand what the job is doing. My question is what it is supposed to be used for?

src/commands/go-test-legacy.yaml
kopiczko: Okay it's answered here https://github.com/giantswarm/architect-orb/pull/75#discussion_r384554291

src/commands/go-test-legacy.yaml
kopiczko: Can you show an example of the build using `go-architect`?

tfussell: https://github.com/giantswarm/apiextensions/blob/103eab523d7831bb8ff6255f3ed7b3d59705820a/.circleci/config.yml

tfussell: https://circleci.com/gh/giantswarm/apiextensions/1324

src/commands/go-test-legacy.yaml
tfussell: I've added the orb+command name prefix to step names now.

src/commands/go-test-legacy.yaml
kopiczko: 👍

src/commands/go-architect.yaml
```diff
@@ -0,0 +1,19 @@
+parameters:
```
kopiczko: I'd like to rename it to have a `-legacy` suffix. At some point we should delete all those legacy jobs.

src/commands/go-cache-save.yaml
```diff
@@ -1,5 +1,10 @@
+parameters:
+  lockfile:
+    type: "string"
+    default: "go.sum"
 steps:
   - save_cache:
-      key: go-build-go-mod-v1-{{ checksum "go.sum" }}
+      key: go-build-go-mod-v1-{{ checksum "<< parameters.lockfile >>" }}
       paths:
+        - "/go/pkg/dep"
```
kopiczko: Can we create a separate legacy commands for saving/restoring cache? I'd like to make the cleanup easier in the future.

tfussell: I tried to reduce code duplication using a default argument and a single command. Happy to add separate legacy commands though.

src/commands/go-architect.yaml
tfussell: Makes sense.

src/commands/go-cache-save.yaml
kopiczko: I see what you mean and I agree with that in general but in this case when we consider those jobs temporary I'd like to make future removal easier.

tfussell: Done.

src/commands/go-architect.yaml
tfussell: Done.

tfussell:
Going to give this one final test before merging.

src/jobs/go-architect-legacy.yaml
```diff
@@ -0,0 +1,12 @@
+description: |
+  This job runs the given command for the current project from a container
+  using the `latest` architect image and ultimately calling `architect <command>`.
+machine: true
```
kopiczko: I wonder if we can use architect executor here and then not use docker in the command.

tfussell: I think I had an issue when I did that at first, but I will confirm.

tfussell: Now I remember. Here's what happened https://circleci.com/gh/giantswarm/apiextensions/1369 The CircleCI runner launches the build in the `architect` container. Then the `architect` binary calls `docker` commands so we have to support docker-in-docker by passing in the docker socket to the `architect` container and I could only get that to work by using `machine: true`.

kopiczko: You have `setup_remote_docker` settings and you can run docker from docker executor. It has limitation of containers not being able to communicate via network but that should be fine for architect. But I'm not big on that. I'm big on removing this job from the orb ASAP (and it's not merged yet :D).

tfussell: I can't figure out an easy way to get the repo onto the remote docker server so I think I'm going to go with the `machine`-based approach for now. As you say, this job is intended to go away in the near future, and I don't want to spend more time on it. And it works as it is.

fiunchinho:
I'm hitting an issue. My docker build fails because the binary was not persisted. Shouldn't there be a parameter like `binary` to persist it so other steps can use it like
```
- persist_to_workspace:	
          root: .	
          paths:	
            - ./azure-operator
```

tfussell:
> I'm hitting an issue. My docker build fails because the binary was not persisted. Shouldn't there be a parameter like `binary` to persist it so other steps can use it like
> 
> ```
> - persist_to_workspace:	
>           root: .	
>           paths:	
>             - ./azure-operator
> ```

I totally neglected this use case since I was doing my testing on a library instead of a binary project. Thanks for bringing it up. I think that your suggested fix will work. This is what the module-based `go-build` does it seems. I'll try to get a fix in shortly.

PR #76, opened 2020-02-21
Add golanglint-ci as linter for Go code

stone-z:
Towards https://github.com/giantswarm/giantswarm/issues/8901 https://github.com/giantswarm/giantswarm/issues/7658

Adds `go-lint` job which can run multiple configurable linters.

## Checklist

### DIFF
--- a/src/jobs/go-lint.yaml
+++ b/src/jobs/go-lint.yaml
@@ -1,0 +1,47 @@
+description: "Runs golanglint-ci against the codebase."
+parameters:
+  disable:
+    default: ""
+    description: "The names of default linting modules to skip."
+    type: "string"
+  disable-all:
+    default: false
+    description: "If set, will disable all linters (except those included by '--enable')."
+    type: "boolean"
+  enable:
+    default: ""
+    description: "The names of additional linting modules to run."
+    type: "string"
+  new-from-rev:
+    default: ""
+    description: "Show only new issues created after git revision REV."
+    type: "string"
+  presets:
+    default: "none"
+    description: "Enable preset list of linters (bugs|complexity|format|performance|style|unused)."
+    type: "enum"
+    enum: ["bugs", "complexity", "format", "performance", "style", "unused", "none"]
+  skip-dirs:
+    default: ""
+    description: "Regular expressions for folder paths which should be skipped."
+    type: "string"
+  skip-files:
+    default: ""
+    description: "Regular expressions for file paths which should be skipped."
+    type: "string"
+  verbose:
+    default: false
+    description: "If set, will print verbose output."
+    type: "boolean"
+executor: go-lint
+steps:
+  - checkout
+  - go-lint:
+      disable: << parameters.disable >>
+      disable-all: << parameters.disable-all >>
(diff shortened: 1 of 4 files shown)


src/jobs/go-lint.yaml
```diff
@@ -0,0 +1,47 @@
...
+    default: "none"
+    description: "Enable preset list of linters (bugs|complexity|format|performance|style|unused)."
+    type: "enum"
+    enum: ["bugs", "complexity", "format", "performance", "style", "unused", "none"]
+  skip-dirs:
+    default: ""
+    description: "Regular expressions for folder paths which should be skipped."
+    type: "string"
+  skip-files:
+    default: ""
```
MarcelMue: Why not have `vendor` here?

src/jobs/go-lint.yaml
```diff
@@ -0,0 +1,47 @@
...
+  enable:
+    default: ""
+    description: "The names of additional linting modules to run."
+    type: "string"
+  new-from-rev:
+    default: ""
+    description: "Show only new issues created after git revision REV."
+    type: "string"
+  presets:
+    default: "none"
```
MarcelMue: Maybe I misunderstand this option, but don't we always want to have at least one?

src/jobs/go-lint.yaml
stone-z: It is already the default

src/jobs/go-lint.yaml
stone-z: There is a default suite, this is if you only want a certain block of linters (these run some which aren't included in the default as well)

src/jobs/go-lint.yaml
MarcelMue: Okay I get it now - the naming `default` is just very confusing to me in that sense :D

stone-z: Yes there are actually Job, Command, and underlying tool defaults to speak of :see_no_evil:

src/jobs/go-lint.yaml
```diff
@@ -0,0 +1,47 @@
...
+  - checkout
+  - go-lint:
+      disable: << parameters.disable >>
+      disable-all: << parameters.disable-all >>
+      enable: << parameters.enable >>
+      new-from-rev: << parameters.new-from-rev >>
+      presets: << parameters.presets >>
+      skip-dirs: << parameters.skip-dirs >>
+      skip-files: << parameters.skip-files >>
+      verbose: << parameters.verbose >>
```
kopiczko: Same as here: https://github.com/giantswarm/architect-orb/pull/77/files#r383888318

I'd like to have the same linter everywhere. If we want to change something between project we should have a good reason for that. We also should have an easy way to include the job into our projects and not force everybody to read all of that. Ideally I just add `- architect/go-lint` and that's it.

kopiczko: Actually we should  have that included in `go-test` and `go-build`.

stone-z:
After feedback, removed most of the arguments in https://github.com/giantswarm/architect-orb/pull/78

PR #77, opened 2020-02-24
Add gitleaks job

stone-z:
Adds gitleaks, the first of two language-agnostic secret scanners. This is primarily for entropy-based matching.

Towards https://github.com/giantswarm/giantswarm/issues/8901 and https://github.com/giantswarm/giantswarm/issues/7657

## Checklist

### DIFF
--- a/src/commands/gitleaks.yaml
+++ b/src/commands/gitleaks.yaml
@@ -16,7 +16,7 @@ parameters:
     description: "Path to a config file to use for gitleaks configuration."
     type: "string"
   debug:
-    description: "If set, will print debug messages."
+    description: "If true, will print debug messages."
     type: "boolean"
   depth:
     description: "Number of commits to audit."
--- a/src/commands/gitleaks.yaml
+++ b/src/commands/gitleaks.yaml
@@ -1,0 +1,113 @@
+parameters:
+  branch:
+    description: "The branch to audit."
+    type: "string"
+  commit:
+    description: "Scan all files changed by this commit (rev. SHA)."
+    type: "string"
+  commit-from:
+    description: "Commit to start audit from (rev. SHA). This is the most recent commit to audit. Default is the latest."
+    type: "string"
+  commit-to:
+    description: "Commit to stop audit (rev. SHA). This is the oldest commit to audit. Default is the root commit. \
+    In CircleCI, you can use << pipeline.git.base_revision >> to scan only since the parent commit."
+    type: "string"
+  config:
+    description: "Path to a config file to use for gitleaks configuration."
+    type: "string"
+  debug:
+    description: "If set, will print debug messages."
+    type: "boolean"
+  depth:
+    description: "Number of commits to audit."
+    type: "integer"
+  files-at-commit:
+    description: "Scan all files as they are in this commit (rev. SHA), even if they did not change."
+    type: "string"
+  pretty:
+    description: "If true, pretty-prints json output (if output is being printed with -v. Default: true)."
+    type: "boolean"
+  redact:
+    description: "If set, redact secrets from log messages and leaks (default: false)."
+    type: "boolean"
+  report:
+    description: "Path to write json leaks file."
+    type: "string"
+  report-format:
+    description: "Format report as json or csv (default: json)."
+    type: "enum"
+    enum: ["json", "csv"]
+  threads:
+    description: "Maximum number of threads gitleaks spawns."
+    type: "integer"
+  timeout:
+    description: "Time allowed per audit. E.g. 10us, 30s, 1m, 1h10m1s."
+    type: "string"
+  verbose:
+    description: "If set, will print verbose output."
+    type: "boolean"
+steps:
+  - run: |
+      BRANCH="<< parameters.branch >>"
+      COMMIT="<< parameters.commit >>"
+      COMMIT_FROM="<< parameters.commit-from >>"
+      COMMIT_TO="<< parameters.commit-to >>"
+      CONFIG="<< parameters.config >>"
+      DEBUG="<< parameters.debug >>"
+      DEPTH="<< parameters.depth >>"
+      FILES_AT_COMMIT="<< parameters.files-at-commit >>"
+      PRETTY="<< parameters.pretty >>"
+      REDACT="<< parameters.redact >>"
+      REPORT="<< parameters.report >>"
+      REPORT_FORMAT="<< parameters.report-format >>"
+      THREADS="<< parameters.threads >>"
+      TIMEOUT="<< parameters.timeout >>"
+      VERBOSE="<< parameters.verbose >>"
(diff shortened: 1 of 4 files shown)


src/commands/gitleaks.yaml
```diff
@@ -0,0 +1,113 @@
...
+      if [ -n "${FILES_AT_COMMIT}" ]; then
+        LINT_ARGS+=" --files-at-commit=${FILES_AT_COMMIT}";
+      fi
+      if [ "${PRETTY}" = "true" ]; then
+        LINT_ARGS+=" --pretty";
+      fi
+      if [ "${REDACT}" = "true" ]; then
+        LINT_ARGS+=" --redact";
+      fi
+      if [ -n "${REPORT}" ]; then
```
MarcelMue: Is this working correctly? There is a lot of bash juggling going on.

src/commands/gitleaks.yaml
```diff
@@ -0,0 +1,113 @@
...
+    type: "string"
+  commit-to:
+    description: "Commit to stop audit (rev. SHA). This is the oldest commit to audit. Default is the root commit. \
+    In CircleCI, you can use << pipeline.git.base_revision >> to scan only since the parent commit."
+    type: "string"
+  config:
+    description: "Path to a config file to use for gitleaks configuration."
+    type: "string"
+  debug:
+    description: "If set, will print debug messages."
```
TheoBrigitte:
```suggestion
    description: "If true, will print debug messages."
```

src/commands/gitleaks.yaml
```diff
@@ -0,0 +1,113 @@
...
+    description: "Number of commits to audit."
+    type: "integer"
+  files-at-commit:
+    description: "Scan all files as they are in this commit (rev. SHA), even if they did not change."
+    type: "string"
+  pretty:
+    description: "If true, pretty-prints json output (if output is being printed with -v. Default: true)."
+    type: "boolean"
+  redact:
+    description: "If set, redact secrets from log messages and leaks (default: false)."
```
TheoBrigitte:
```suggestion
    description: "If true, redact secrets from log messages and leaks (default: false)."
```

src/commands/gitleaks.yaml
```diff
@@ -0,0 +1,113 @@
...
+    type: "enum"
+    enum: ["json", "csv"]
+  threads:
+    description: "Maximum number of threads gitleaks spawns."
+    type: "integer"
+  timeout:
+    description: "Time allowed per audit. E.g. 10us, 30s, 1m, 1h10m1s."
+    type: "string"
+  verbose:
+    description: "If set, will print verbose output."
```
TheoBrigitte:
```suggestion
    description: "If true, will print verbose output."
```

src/jobs/gitleaks.yaml
```diff
@@ -0,0 +1,84 @@
...
+    description: "Commit to stop audit (rev. SHA). This is the oldest commit to audit. Default is the root commit. \
+    In CircleCI, you can use << pipeline.git.base_revision >> to scan only since the parent commit."
+    type: "string"
+  config:
+    default: ""
+    description: "Path to a config file to use for gitleaks configuration."
+    type: "string"
+  debug:
+    default: false
+    description: "If set, will print debug messages."
```
TheoBrigitte:
```suggestion
    description: "If true, will print debug messages."
```

src/jobs/gitleaks.yaml
```diff
@@ -0,0 +1,84 @@
...
+    default: ""
+    description: "Scan all files as they are in this commit (rev. SHA), even if they did not change."
+    type: "string"
+  pretty:
+    default: true
+    description: "If true, pretty-prints json output (default: true)."
+    type: "boolean"
+  redact:
+    default: false
+    description: "If set, redact secrets from log messages and leaks (default: false)."
```
TheoBrigitte:
```suggestion
    description: "If true, redact secrets from log messages and leaks (default: false)."
```

src/jobs/gitleaks.yaml
```diff
@@ -0,0 +1,84 @@
...
+    default: 0
+    description: "Maximum number of threads gitleaks spawns (default: ALL - 1)."
+    type: "integer"
+  timeout:
+    default: ""
+    description: "Time allowed per audit. E.g. 10us, 30s, 1m, 1h10m1s."
+    type: "string"
+  verbose:
+    default: false
+    description: "If set, will print verbose output."
```
TheoBrigitte:
```suggestion
    description: "If true, will print verbose output."
```

stone-z:
> There is a lot of bash juggling going on

> that's a lot of default values

Agreed, but since this check will potentially cause a lot of noise, I wanted to expose as much configuration as I think we'll need right off the bat so we can tune it for each repository.

Unfortunately CircleCI won't allow an optional parameter without a default value, and there isn't a nice way to ask CircleCI to line up these parameters; even the [official CircleCI k8s orb](https://github.com/CircleCI-Public/kubernetes-orb/blob/master/src/commands/delete-resource.yml#L108) does this. But since there isn't any real logic there I'm ok with using this until there is a better way or a reason to change it.

src/jobs/gitleaks.yaml
```diff
@@ -0,0 +1,84 @@
...
+      debug: << parameters.debug >>
+      depth: << parameters.depth >>
+      files-at-commit: << parameters.files-at-commit >>
+      pretty: << parameters.pretty >>
+      redact: << parameters.redact >>
+      report: << parameters.report >>
+      report-format: << parameters.report-format >>
+      threads: << parameters.threads >>
+      timeout: << parameters.timeout >>
+      verbose: << parameters.verbose >>
```
kopiczko: Why do we expose all those parameters? I'd strongly prefer to have all or almost all baked-in in the orb. Parameters can be exposed as they are needed. Otherwise we allow more deviation between the projects than necessary.

src/commands/gitleaks.yaml
```diff
@@ -0,0 +1,113 @@
...
+        LINT_ARGS+=" --threads=${THREADS}";
+      fi
+      if [ -n "${TIMEOUT}" ]; then
+        LINT_ARGS+=" --timeout=${TIMEOUT}";
+      fi
+      if [ "${VERBOSE}" = "true" ]; then
+        LINT_ARGS+=" --verbose";
+      fi
+      echo "Running gitleaks $LINT_ARGS -r ./"
+      gitleaks $LINT_ARGS -r ./
```
kopiczko: From https://github.com/giantswarm/architect-orb#architect-orb:

> Having one line binary call on each job step - to make the outputs grouped and visible and keeping build configurations sane. Good example is package-and-push command.

stone-z:
After feedback, removed parameters in https://github.com/giantswarm/architect-orb/pull/79

PR #78, opened 2020-02-26
Remove params

stone-z:
new-from-rev is suggested so we don't have to re-lint the whole repo on every change, but this could also be removed. The name wouldn't be my first choice, but I want to keep consistency with the upstream tool

Towards https://github.com/giantswarm/giantswarm/issues/8901 and https://github.com/giantswarm/giantswarm/issues/7658

## Checklist

~~- [ ] Update changelog in CHANGELOG.md.~~ - no change

### DIFF
--- a/src/commands/go-lint.yaml
+++ b/src/commands/go-lint.yaml
@@ -1,64 +1,15 @@
 parameters:
-  # Many additional parameters are supported by golanglint-ci.
-  disable:
-    description: "The names of default linting modules to skip."
-    type: "string"
-  disable-all:
-    description: "If set, will disable all linters (except those included by '--enable')."
-    type: "boolean"
-  enable:
-    description: "The names of additional linting modules to run."
-    type: "string"
   new-from-rev:
     description: "Show only new issues created after git revision REV."
     type: "string"
-  presets:
-    description: "Enable preset list of linters (bugs|complexity|format|performance|style|unused)."
-    type: "enum"
-    enum: ["bugs", "complexity", "format", "performance", "style", "unused", "none"]
-  skip-dirs:
-    description: "Regular expressions for folder paths which should be skipped."
-    type: "string"
-  skip-files:
-    description: "Regular expressions for file paths which should be skipped."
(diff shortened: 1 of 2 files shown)


src/commands/go-lint.yaml
```diff
@@ -1,64 +1,15 @@
...
-    type: "string"
-  disable-all:
-    description: "If set, will disable all linters (except those included by '--enable')."
-    type: "boolean"
-  enable:
-    description: "The names of additional linting modules to run."
-    type: "string"
   new-from-rev:
     description: "Show only new issues created after git revision REV."
     type: "string"
```
kopiczko: Why would we keep that option open? Is the linter too slow to check all files every single time?

stone-z: No the speed is fine, the idea was just to provide some way to scope the check. For example in the case of a feature branch, we might be interested in only blocking a build for new issues. This could be removed, then we'll just always get output for the whole history

kopiczko: I think we should make the decision here and not in every single repo build config. For me either is fine.

stone-z: Fine, I'd rather have everything visible in that case. Removed

src/commands/go-lint.yaml
```diff
@@ -1,64 +1,5 @@
...
-      if [ -n "${SKIP_FILES}" ]; then
-        LINT_ARGS+=" --skip-files ${SKIP_FILES}";
-      fi
-      if [ "${VERBOSE}" = "true" ]; then
-        LINT_ARGS+=" -v";
-      fi
-      echo "Running: golangci-lint run $LINT_ARGS"
-      golangci-lint run $LINT_ARGS
+      # Enable some additional linters
+      echo "Running: golangci-lint run -E gosec -E goconst -E unparam"
```
kopiczko:
```suggestion
```

stone-z:
Will add to go-test job in new PR, for my sanity

PR #79, opened 2020-02-26
Remove all parameters except config

stone-z:
I think this would be minimal for us to work with. Keeps the ability to whitelist files and limit the scope to a single commit or changes since the last rev

Toward https://github.com/giantswarm/giantswarm/issues/8901

## Checklist

~~- [ ] Update changelog in CHANGELOG.md.~~ - no change

### DIFF
--- a/src/commands/gitleaks.yaml
+++ b/src/commands/gitleaks.yaml
@@ -1,28 +1,13 @@
 parameters:
-  commit-to:
-    description: "Commit to stop audit (rev. SHA). This is the oldest commit to audit. Default is the root commit. \
-    In CircleCI, you can use << pipeline.git.base_revision >> to scan only since the parent commit."
-    type: "string"
   config:
-    description: "Path to a config file to use for gitleaks configuration."
-    type: "string"
-  files-at-commit:
-    description: "Scan all files as they are in this commit (rev. SHA)."
+    description: "(Optional) Path to a config file to use for gitleaks configuration."
     type: "string"
 steps:
   - run: |
-      COMMIT_TO="<< parameters.commit-to >>"
       CONFIG="<< parameters.config >>"
-      FILES_AT_COMMIT="<< parameters.files-at-commit >>"
       LINT_ARGS=" "
-      if [ -n "${COMMIT_TO}" ]; then
-        LINT_ARGS+=" --commit-to=${COMMIT_TO}";
-      fi
       if [ -n "${CONFIG}" ]; then
         LINT_ARGS+=" --config=${CONFIG}";
-      fi
-      if [ -n "${FILES_AT_COMMIT}" ]; then
-        LINT_ARGS+=" --files-at-commit=${FILES_AT_COMMIT}";
       fi
       echo "Running gitleaks $LINT_ARGS --redact -v --pretty -r ./"
       gitleaks $LINT_ARGS --redact -v --pretty -r ./
--- a/src/commands/gitleaks.yaml
+++ b/src/commands/gitleaks.yaml
@@ -1,13 +1,4 @@
 parameters:
-  branch:
-    description: "The branch to audit."
-    type: "string"
-  commit:
-    description: "Scan all files changed by this commit (rev. SHA)."
-    type: "string"
-  commit-from:
-    description: "Commit to start audit from (rev. SHA). This is the most recent commit to audit. Default is the latest."
-    type: "string"
   commit-to:
     description: "Commit to stop audit (rev. SHA). This is the oldest commit to audit. Default is the root commit. \
     In CircleCI, you can use << pipeline.git.base_revision >> to scan only since the parent commit."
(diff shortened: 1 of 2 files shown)


src/commands/gitleaks.yaml
```diff
@@ -1,113 +1,28 @@
...
-  commit:
-    description: "Scan all files changed by this commit (rev. SHA)."
-    type: "string"
-  commit-from:
-    description: "Commit to start audit from (rev. SHA). This is the most recent commit to audit. Default is the latest."
-    type: "string"
   commit-to:
     description: "Commit to stop audit (rev. SHA). This is the oldest commit to audit. Default is the root commit. \
     In CircleCI, you can use << pipeline.git.base_revision >> to scan only since the parent commit."
     type: "string"
```
kopiczko: Why not using `<< pipeline.git.base_revision >>` and removing this param?

src/commands/gitleaks.yaml
```diff
@@ -1,113 +1,28 @@
...
   config:
     description: "Path to a config file to use for gitleaks configuration."
     type: "string"
```
kopiczko: Why should we change the config between builds?

src/commands/gitleaks.yaml
```diff
@@ -1,113 +1,28 @@
...
-  debug:
-    description: "If true, will print debug messages."
-    type: "boolean"
-  depth:
-    description: "Number of commits to audit."
-    type: "integer"
   files-at-commit:
```
kopiczko: My understanding is this flag allows to specify commit sha to check. What is the use case to check only fixed commit in the CI setup?

src/commands/gitleaks.yaml
stone-z: You can't specify a pipeline variable inside the actual orb, only from inside a workflow. So this has to be there in order to pass the base revision from the workflow to the orb. 

If you're curious, CircleCI says they're working on a replacement, but they don't support pipeline variables in orbs because it would make builds un-reproduceable since subsequent runs of a job with the same configuration might result in different orb behavior.

src/commands/gitleaks.yaml
stone-z: For example, different rules for master vs branch builds

src/commands/gitleaks.yaml
stone-z: I'm anticipating noise. I might not want to hear about every high-entropy thing from all time when I make some small change

src/commands/gitleaks.yaml
kopiczko: Per branch changes we should be able to cover from inside the orb IMO.

src/commands/gitleaks.yaml
kopiczko: I see. That sucks balls. Let's please add an example to the README.md.

src/commands/gitleaks.yaml
kopiczko: I don't get that. Does that allow to specify files to check or commit to check?

stone-z: It is a commit hash, but it scans all files at the state of that commit, not just the files which were modified. The default behavior would show results for every commit ever. I was thinking this would be more applicable for pushes to branches and master can scan the whole history. That way it isn't noisy for people during daily tasks and we can work on tuning the rules for master

kopiczko: I wonder:

- What is the workflow to change that parameter. Should we do this in every PR? That doesn't sound very efficient.
- How useful is having those full reports in the master branch? IMO the job is useful only if it fails when something is wrong otherwise I can guarantee you it will be ignored in 99% cases.

src/commands/gitleaks.yaml
stone-z: I'm still convinced we should expose some way to tune this, but I guess the alternative is assuming every repository will use a config file with the same path. This path must exist, so the orb could detect if it exists and change the command based on whether the file is there or not. Would that be better?

src/commands/gitleaks.yaml
stone-z: I will remove this as well. In the absence of using different configs for branches or different logic this isn't useful.

src/commands/gitleaks.yaml
stone-z: I will remove this as well. In the absence of using different configs for branches or different logic this isn't useful.

src/commands/gitleaks.yaml
kopiczko: We talked about this. The `config` param stays. We want to at least be able to whitelist files.

src/commands/gitleaks.yaml
```diff
@@ -1,113 +1,13 @@
...
-      fi
-      if [ -n "${TIMEOUT}" ]; then
-        LINT_ARGS+=" --timeout=${TIMEOUT}";
-      fi
-      if [ "${VERBOSE}" = "true" ]; then
-        LINT_ARGS+=" --verbose";
-      fi
-      echo "Running gitleaks $LINT_ARGS -r ./"
-      gitleaks $LINT_ARGS -r ./
+      echo "Running gitleaks $LINT_ARGS --redact -v --pretty -r ./"
```
kopiczko: CirlcleCI will print the command anyway.

```suggestion
```

PR #80, opened 2020-02-28
Add feature to test Helm charts

yasn77:
This was originally opened #65 but that was closed. Now architect Docker image contains chart testing, reopening.

Using chart testing (https://github.com/helm/chart-testing), add job to lint
Helm charts.

Chart testing is pretty extensive and we can easily add other tests using it.

IMHO this is the first step, next would be to remove the executor by adding `ct`
to architect docker image. After that look to extend architect command itself to
either call ct or use ct as a module (not sure if that is possible).

## Checklist

- [ ] Update changelog in CHANGELOG.md.


src/jobs/helm-lint.yaml
```diff
@@ -0,0 +1,24 @@
+---
```
kopiczko: Do we need to make that a separate job? Can we include this in existing one? Where do you want to use it?

yasn77: There isn't really any existing job where this can go... The current Helm related jobs will package and publish Helm chart. The idea of this job is to do linting (and in future testing) of the Helm chart before we publish. So for example when PR is open, this will lint the Helm chart before it gets merged.

kopiczko: How about making this part of push-to-app-catalog job then? We should have it everywhere.

yasn77: The way it's currently configured means that if I add it to push-to-app-catalog, it will only run if a tag is created (or w want to publish the Helm package). This isn't really what I think we want, we want to test before we commit to trying to create a package.

kopiczko: push-to-app-catalog runs on every commit and tag. push-to-app-collection runs only on tag.

I'd say we want to test before pushing the chart to the catalog. Before that the chart is only built locally in the CI. Having the command included in existing job ensures that it's run once the architect orb version is bumped in the configuration.

yasn77: Maybe I've misunderstood the purpose of the jobs, but looking at the Kong config (done before I was around) and Nginx app, both only run `push-to-app-catalog` for `^v.*` tags:

https://github.com/giantswarm/nginx-ingress-controller-app/blob/master/.circleci/config.yml#L92
https://github.com/giantswarm/kong-app/blob/master/.circleci/config.yml#L14

I agree that we want to use these tests are run before pushing, and putting them in the existing job is good. But I also think that we should run these tests for PR and not push. Would it right to create the jobs (in this PR) and also include it in existing jobs?

kopiczko: This is CircleCI weirdness. Jobs are by default run for all branches and no tags. So to enable building _also_ for tags you have to add this filter. To build only on tags you'd have to do this:

https://github.com/giantswarm/aws-operator/blob/1e62a0b7dde996640f9f01832526ce284ee847ad/.circleci/config.yml#L118-L124

yasn77: Ah ok, that makes sense now :)

I am cool to add this to existing job, I'll update branch and PR. Thanks

yasn77: @kopiczko Implemented your suggestion, I have removed `helm-lint` job and incorporated the steps into `pust-to-app-catalog`. The change I had to make was to move template step, as it needs to run early for linting to work. Thanks

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Introduce `helm-lint` which uses Helm chart testing to lint Helm Charts
```
kopiczko: I don't consider commands to be real interface of the orb. Let's mention changes to the job instead which is important from the user perspective.

```suggestion
- Introduce Helm Chart testing and linting in `push-to-app-catalog` job.
```

src/jobs/push-to-app-catalog.yaml
```diff
@@ -9,10 +9,27 @@ parameters:
...
+      Chart Testing config file
+      (see: https://github.com/helm/chart-testing#configuration)
+    type: string
+    default: ""
+  chart_testing_additional_args:
+    description: >-
+      Additonal command line arguments
+      (ignored if config_file is set)
+    type: string
+    default: ""
```
kopiczko: A few points here:

- Are you really sure we need those parameters and can't have sane defaults everywhere instead? It's easy to expose new parameters but hiding them is a breaking change.
- If they are really needed: how about replacing this with path to actual files holding the configuration in the repo?
- Let's replace `_` with `-`. I know this is inconsistent across the repo but the later is what we should have everywhere.

yasn77: > * Are you really sure we need those parameters and can't have sane defaults everywhere instead? It's easy to expose new parameters but hiding them is a breaking change.
> * If they are really needed: how about replacing this with path to actual files holding the configuration in the repo?

On these points, I have removed `additional-args`, but kept the config file... The config file can/will exist in the repo of the chart and it's up to the chart maintainer to enable disable things as part of linting/testing. If config file is omitted, then it will run using defaults/

> * Let's replace `_` with `-`. I know this is inconsistent across the repo but the later is what we should have everywhere.

Done

Thanks

kopiczko: How about having our custom config-we-want-to-use-everywhere defined here? Are there configurations that _have_ to be changed between repositories?

yasn77: Personally, I am not sure. Because it means that if an App maintainer needs to deviate from the set values, they will need to raise a PR here to get that changed. I don't think that gains us anything.

I would like to at least keep the option of using a config file.

kopiczko: Deviation means lack of consistency which we value a lot. If something needs to be adjusted we should rather do that for all the apps (which is in the orb itself). Unless there is a good/known reason to configure it differently per app. Still if we can't anticipate it now and it turns out that we need per app adjustment we can expose a parameter later on.

yasn77: Cool... I have simplified further. The lint doesn't take any parameters and uses defaults (except checking for maintainers which don't really apply)

cokiengchiara:
Needs to be documented.

cokiengchiara:
Yasser updated PR to update docs.

PR #81, opened 2020-03-02
Add integration-test job

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/4606

Spec: https://github.com/giantswarm/giantswarm/blob/5bc660c2e3bcc107b718bb8b3ebcdab586007ed0/sigs/releng/specs/revamp_e2e_spec.md

## Checklist


kopiczko:
I added my comments here https://github.com/giantswarm/app-operator/pull/197 couple minutes ago. Should I move them here?

rossf7:
Eek I submitted this a bit early. Going to address the feedback in the inline PR first. https://github.com/giantswarm/app-operator/pull/197

Sorry for the noise. I'll post again when this is ready for review.

src/commands/integration-test-create-cluster.yaml
```diff
@@ -0,0 +1,16 @@
...
+      steps:
+        - run:
+            name: "Prepare default kind config"
+            command: |
+              echo "kind: Cluster" >> << parameters.kind-config >>
+              echo "apiVersion: kind.sigs.k8s.io/v1alpha3" >> << parameters.kind-config >>
+  - run:
+      name: "Create cluster"
+      command: |
+        kind create cluster --config  << parameters.kind-config >>  --image quay.io/giantswarm/kind-node:$KUBERNETES_VERSION
```
kopiczko: When << parameters.kind-config >> is empty `echo "kind: Cluster" >> << parameters.kind-config >>` would write to file "".

```suggestion
  - when:
      condition: << parameters.kind-config >>
      steps:
        - run:
            name: "Move kind config"
            command: |
                cp << parameters.kind-config >> .kind-config
  - unless:
      condition: << parameters.kind-config >>
      steps:
        - run:
            name: "Prepare default kind config"
            command: |
                printf "kind: Cluster\napiVersion: kind.sigs.k8s.io/v1alpha3\n" .kind-config
  - run:
      name: "Create cluster"
      command: |
          kind create cluster --config  .kind-config  --image quay.io/giantswarm/kind-node:$KUBERNETES_VERSION
```

src/commands/integration-test-go-test.yaml
```diff
@@ -0,0 +1,21 @@
+parameters:
+  env-file:
+    type: string
+  test-dir:
+    type: string
+steps:
+  - run:
+      name: "Process .env file"
+      command: |
+        echo '<< parameters.env-file >>' | sed -e 's/^\(.\)/export \1/' | tee .env
```
kopiczko: It would be good to surround this with `when:`

src/commands/integration-test-go-test.yaml
```diff
@@ -0,0 +1,21 @@
...
+      # TODO remove this part after switch to Go modules.
+      name: "Move code to $GOPATH"
+      command: |
+        mkdir -p $GOPATH/src/github.com/${CIRCLE_PROJECT_USERNAME} && mv ~/project $GOPATH/src/github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}
+  - run:
+      name: "Run test"
+      command: |
+        cd $GOPATH/src/github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}
+        source .env
+        CGO_ENABLED=0 E2E_KUBECONFIG=~/.kube/config go test -tags k8srequired -timeout 20m -v << parameters.test-dir >>
```
kopiczko: Can you apply the suggestion I made in the other repo?

src/commands/integration-test-create-cluster.yaml
```diff
@@ -0,0 +1,16 @@
+parameters:
+  kind-config:
+    type: string
+steps:
+  - unless:
+      condition: << parameters.kind-config >>
+      steps:
+        - run:
+            name: "Prepare default kind config"
```
kopiczko: Let's please add the command prefix to the job names like here https://github.com/giantswarm/architect-orb/blob/29f86bd77f1ad5de0fca201f85148fb1d853af18/src/commands/go-test-legacy.yaml#L3
That makes it easier to realise what comes from where when all that is concatenated into a single yaml.

src/commands/integration-test-install-go.yaml
```diff
@@ -0,0 +1,33 @@
+steps:
```
kopiczko: I'd rename it to `machine-install-go` - I can see this command being used in different jobs as well.

src/commands/integration-test-create-cluster.yaml
rossf7: Yes I tried to remove the file copy but its needed. Going with this, thx!

src/commands/integration-test-create-cluster.yaml
```diff
@@ -0,0 +1,22 @@
...
+            name: "architect/integration-test-create-cluster: Move kind config"
+            command: |
+                cp << parameters.kind-config >> .kind-config
+  - unless:
+      condition: << parameters.kind-config >>
+      steps:
+        - run:
+            name: "architect/integration-test-create-cluster: Prepare default kind config"
+            command: |
+                printf "kind: Cluster\napiVersion: kind.sigs.k8s.io/v1alpha3\n" .kind-config
```
kopiczko:
```suggestion
                printf "kind: Cluster\napiVersion: kind.sigs.k8s.io/v1alpha3\n" | tee .kind-config
```

src/commands/integration-test-create-cluster.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - run:
+      name: "architect/integration-test-create-cluster: Create cluster"
+      command: |
+          kind create cluster --config  .kind-config  --image quay.io/giantswarm/kind-node:$KUBERNETES_VERSION
```
kopiczko:
```suggestion
          kind create cluster --config .kind-config --image quay.io/giantswarm/kind-node:$KUBERNETES_VERSION
```

src/jobs/integration-test.yaml
```diff
@@ -0,0 +1,43 @@
...
+  env-file:
+    default: ".env"
+    description: "File of environment variables to set."
+    type: string
+  kind-config:
+    default: ".kind-config"
+    description: "Path to kind config file."
+    type: string
+  setup-script:
+    default: "setup.sh"
```
kopiczko: This should be empty by default so we don't trigger setup commands with non-existing path by default.

```suggestion
    default: ""
```

src/jobs/integration-test.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+  Tests can be customized by providing a KIND config file or a setup script
+  which is run before the test is executed.
+environment:
+  GOPATH: /go
+  KUBERNETES_VERSION: v1.16.3
+executor: machine
+parameters:
+  env-file:
+    default: ".env"
```
kopiczko: I guess empty is a sane default here. We don't need to always have env file.

```suggestion
    default: ""
```

src/jobs/integration-test.yaml
```diff
@@ -0,0 +1,43 @@
...
+    description: "File of environment variables to set."
+    type: string
+  kind-config:
+    default: ".kind-config"
```
kopiczko: Same here.

```suggestion
    default: ""
```

src/jobs/integration-test.yaml
rossf7: Yes good shout. Fixed.

src/jobs/integration-test.yaml
```diff
@@ -0,0 +1,43 @@
...
+  Runs an integration test by creating a KIND cluster and executing it as a Go
+  test.
+
+  Uses the machine executor and replaces the Go version with a version
+  controlled by us since the preinstalled Go is very outdated.
+
+  Tests can be customized by providing a KIND config file or a setup script
+  which is run before the test is executed.
+environment:
+  GOPATH: /go
```
kopiczko: Why do we need to change the GOPATH here? It's already set up there https://github.com/giantswarm/architect-orb/pull/81/files#diff-af2873c4ca693176b903875216a77aa9R22-R25

src/jobs/integration-test.yaml
```diff
@@ -0,0 +1,43 @@
...
+  KUBERNETES_VERSION: v1.16.3
```
kopiczko: Where is this used?

src/commands/integration-test-export-logs.yaml
```diff
@@ -0,0 +1,9 @@
+steps:
+  - run:
+      name: "architect/integration-test-export-logs: Export cluster logs"
+      command: |
+        kind export logs ./kind/logs
+      when: always
```
kopiczko: I think it would be good to move `when: always` part to the job to make it more visible at the high level.

src/jobs/integration-test.yaml
```diff
@@ -0,0 +1,43 @@
...
+    description: "Path to setup script with commands to execute before each test."
+    type: string
+  test-dir:
+    description: "Tested package directory."
+    type: string
+steps:
+  - checkout
+  - attach_workspace:
+      at: .
+  - machine-install-go
```
kopiczko: I think restoring build cache would be beneficial here to save some time and github quota 
```suggestion
  - go-cache-restore
  - machine-install-go
```

src/commands/integration-test-setup.yaml
```diff
@@ -0,0 +1,15 @@
...
+      condition: << parameters.setup-script >>
+      steps:
+        - run:
+            name: "architect/integration-test-setup: Output test setup commands"
+            command: |
+              cat << parameters.setup-script >>
+        - run:
+            name: "architect/integration-test-setup: Run test setup commands"
+            command: |
+              sh << parameters.setup-script >>
```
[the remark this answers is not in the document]
rossf7: Yes good point. I'll update.

src/jobs/integration-test.yaml
rossf7: It's used here. https://github.com/giantswarm/architect-orb/blob/fcdc0e0f2834bfdc88a7534e2abbacd6211c1dc2/src/commands/integration-test-create-cluster.yaml#L22

IIRC I did this following a request from @MarcelMue that the k8s version be explicit and obvious.

Using an env var means you can override. I think this is useful for testing new k8s releases before we change the default.

src/jobs/integration-test.yaml
rossf7: I think its a leftover from previous PRs. I'll remove.

src/jobs/integration-test.yaml
```diff
@@ -0,0 +1,43 @@
...
+  - integration-test-install-tools
+  - integration-test-create-cluster:
+      kind-config: << parameters.kind-config >>
+  - integration-test-setup:
+      setup-script: << parameters.setup-script >>
+  - integration-test-go-test:
+      env-file: << parameters.env-file >>
+      test-dir: << parameters.test-dir >>
+  - integration-test-export-logs
```
kopiczko: How about adding `when: always` here instead of the command itself? I think it makes things clearer from the high level veiw.

```suggestion
  - integration-test-export-logs
    when: always
```

src/jobs/integration-test.yaml
kopiczko: If we need to be able to override that then let's make this a parameter. I'd still like to avoid being able to do that because I'm afraid this can be abused. I can disagree and commit though.

src/jobs/integration-test.yaml
rossf7: Yes makes sense. Done.

rossf7: Nah linter says no. I'll get it sorted.

rossf7:
@kopiczko Thanks for all your help with this! <3

Removing the go env var broke it but I think I know how to fix. I get that sorted and then merge this.

rossf7:
Tested in https://github.com/giantswarm/app-operator/pull/211. Going with this.

I still need to get the build cache to work. I do that as follow up because this PR was big and noisy.

PR #87, opened 2020-03-05
Run go-lint as part of go-test command for legacy and modules

stone-z:
Runs golangci-lint as part of the existing go-test and go-test-legacy commands instead of setting up separate environment and using separate executor.

## Checklist

### DIFF
--- a/src/executors/architect.yaml
+++ b/src/executors/architect.yaml
@@ -1,3 +1,3 @@
 docker:
   - entrypoint: /bin/bash
-    image: quay.io/giantswarm/architect:0.0.0-0670068376661ce04ff6f6285bb1f740d64120c0
+    image: quay.io/giantswarm/architect:latest
--- a/src/executors/architect.yaml
+++ b/src/executors/architect.yaml
@@ -1,3 +1,3 @@
 docker:
   - entrypoint: /bin/bash
-    image: quay.io/giantswarm/architect:latest
+    image: quay.io/giantswarm/architect:0.0.0-0670068376661ce04ff6f6285bb1f740d64120c0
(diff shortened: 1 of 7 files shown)


stone-z:
Tested on ignition-operator (go modules) here: https://circleci.com/gh/giantswarm/ignition-operator/309
Tetsed on apiextensions (dep) here: https://circleci.com/gh/giantswarm/apiextensions/1457

corest:
Do we have this linter mentioned somewhere in https://github.com/giantswarm/fmt ? As you would need to have it installed and used in your dev env before it is added into the test and making tests red

stone-z:
> Do we have this linter mentioned somewhere in https://github.com/giantswarm/fmt ? As you would need to have it installed and used in your dev env before it is added into the test and making tests red

Good point. I will add a reference there. Will also make it a point to announce the new behavior so people anticipate it.

[ ] add tool to fmt
[ ] announce linting behavior and reference tool documentation

src/executors/architect.yaml
```diff
@@ -1,3 +1,3 @@
 docker:
   - entrypoint: /bin/bash
-    image: quay.io/giantswarm/architect:latest
+    image: quay.io/giantswarm/architect:0.0.0-0670068376661ce04ff6f6285bb1f740d64120c0
```
MarcelMue: Revert this when merged?

stone-z:
`golangci-lint` offers a `gofmt` linter as well. This is already run as part of the test -- any preferences on running it as-is or moving it to `golangci-lint`?

src/executors/architect.yaml
stone-z: Reverted

PR #88, opened 2020-03-05
Release v0.8.0

tfussell:
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
 ## [0.8.0] - 2020-03-05
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -142,7 +142,8 @@ which enable `dep` dependencies to be cached in jobs as long as `Gopkg.lock` doe
 
 - Add push-to-app-catalog job.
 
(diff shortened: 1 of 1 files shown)


tfussell:
There's a weird failure on the `lint` step. `Either git or ssh (required by git to clone through SSH) is not installed in the image.` Found a related issue, but I'm not sure why this is only showing up on a tag https://github.com/CircleCI-Public/shellcheck-orb/issues/9

kopiczko:
> but I'm not sure why this is only showing up on a tag

What do you mean?


In general I feel like this some github<->ci weirdness. You can try to rerun whole workflow. `lint` would pass anyway so if this is one-off quirk I think it's ok to ignore that.

tfussell:
> > but I'm not sure why this is only showing up on a tag
> 
> What do you mean?
> 
> In general I feel like this some github<->ci weirdness. You can try to rerun whole workflow. `lint` would pass anyway so if this is one-off quirk I think it's ok to ignore that.

These are the same code https://circleci.com/gh/giantswarm/architect-orb/1320 https://circleci.com/gh/giantswarm/architect-orb/1325 but the latter one was triggered by my tagging `v0.8.0`. Maybe it's because I should have merged and then tagged rather than tagging the commit on this branch. I will delete the tag to find out.

tfussell:
Looks like it's actually this issue https://discuss.circleci.com/t/checkout-fails-on-workflow-triggered-by-annotated-git-tag/27272 Possible fixes are retagging/editing the image the linter to include a newer version of git or cloning manually.

tfussell:
I pushed a tag temporarily just to ensure that the issue was fixed (didn't actually publish to the orb registry). The job worked that time https://circleci.com/gh/giantswarm/architect-orb/1355 I think this is ready to go now.

PR #92, opened 2020-03-09
helm-chart-template: verify chart version in tagged operator builds

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/8635

Tests:
- operator and chart versions don't match https://circleci.com/gh/giantswarm/pawel-test/19
- operator and tag versions don't match https://circleci.com/gh/giantswarm/pawel-test/20
- operator, chart and tag versions match (build fails but the helm-chart-template passes with validation) https://circleci.com/gh/giantswarm/pawel-test/21 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,9 +13,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Support for optionally attaching the persisted workspace in the `push-to-app-catalog` job.
 - Introduce Helm Chart testing and linting in `push-to-app-catalog` job.
-- Add changelog-lint job.
+- Add `changelog-lint` job.
 - Introduce `integration-test` job for running `Go` integration tests in a `KIND`
   cluster.
+- Verify chart, operator and tag versions while packing helm chart on tagged
+  operator build in `push-to-app-collection` job.
 
 ### Changed
 
--- a/src/commands/helm-chart-template.yaml
+++ b/src/commands/helm-chart-template.yaml
@@ -5,7 +5,8 @@ parameters:
 steps:
   - run:
       name: "architect/helm-chart-template: Template chart"
-      command: architect helm template --dir ./helm/<<parameters.chart>>
+      command: |
+          architect helm template --dir ./helm/<<parameters.chart>>
   - run:
       name: "architect/helm-chart-template: Check if build is triggered by a tagged operator"
       command: |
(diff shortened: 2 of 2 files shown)


kopiczko:
@oponder I understand your request but I don't want to pollute and possibly break the job to display nicer messages (exit codes are crucial here and `echo` exits with `0`) . I think the diff output is pretty decent (that's why I decided to use diff):

```
diff --git a/.build_operator_version b/.build_chart_version
index 69112e2..e88320d 100644
--- a/.build_operator_version
+++ b/.build_chart_version
@@ -1 +1 @@
-8.1.2-dev
+20.0.0
```

Screenshot 2020-03-09 at 16 29 54

JosephSalisbury:
Yeah, I'm kind of agreeing with @oponder here. I have no idea what that diff means :D

kopiczko:
I think this is fairly simple to "decode":

```
diff --git a/.build_operator_version b/.build_chart_version
index 69112e2..e88320d 100644
--- a/.build_operator_version
+++ b/.build_chart_version
@@ -1 +1 @@
-8.1.2-dev
+20.0.0
```

operator version is 8.1.2-dev and chart version is 20.0.0 and they differ.

I thought about implementation and I think w/o some black magic and using things like `eval` it will be difficult to achieve. I don't know how to carry over the return status and exit with it after `|| echo` with within the same command given that we have `-eo pipefail` bash settings.

Checking return status after running `git diff` doesn't make sense if it fails because the execution will never reach there.

Alternative would be writing an architect binary command which would do the checks but then we lose the explicit nature of the thing and it will be way more time consuming.

I can print debugging cues before executing `git diff` if that's helpful. Otherwise please provide some suggestions to the code because I don't know how to solve this problem. I still think having a check w/o nice messages is more valuable than having broken releases.

JosephSalisbury:
I think some debug cue before the `git diff` would be great.

e.g: "Checking operator version and chart version are the same - this is required in tagged builds" would be grand

kopiczko:
> e.g: "Checking operator version and chart version are the same - this is required in tagged builds" would be grand

Isn't that basically the same thing?
https://github.com/giantswarm/architect-orb/pull/92/files#diff-1a2d1efa4c648bf0b3cfa2fdf8dcaa7dR32

JosephSalisbury:
Oh wow, I really didn't even see that. That feels easy to miss :D

kopiczko:
> Oh wow, I really didn't even see that. That feels easy to miss :D

I don't have impact on the circle UI. I think step names are pretty visible or at least one can get used to that. See https://circleci.com/gh/giantswarm/pawel-test/19.

Can I please get either approval or a suggestion? I feel stuck.

JosephSalisbury:
I think I get the constraints now. @oponder you good?

TheoBrigitte:
Last minute suggestion: how about having this check as a separate job named `ensure operator version and chart version are the same` which only run for tagged build. Then there is no question whether version have to be the same, and it will also avoid all those ifs.

kopiczko:
> Last minute suggestion: how about having this check as a separate job named `ensure operator version and chart version are the same` which only run for tagged build. Then there is no question whether version have to be the same, and it will also avoid all those ifs.

That's a good call. My preference though is to have it in the existing job. That way we don't need to change the existing CI configs. This is very important to have that included in the build because it prevents from having bad releases and packaging the chart is required for having a release. So even if it's a bit uglier I still prefer to have it that way.

PR #99, opened 2020-03-10
Options to disable and make helm lint less strict

yasn77:
There are cases where existing apps (not managed apps) will have a minimal Chart
file. In these cases, give the developers an option to make the check less
strict or completely disable.

Disabling might be useful where templates deliberately reference empty or non
existing values (in cases where values are taken from installations repo). It is
possible to create dummy values, but I think giving the option to disable
creates less of an impediment to development.

In both cases (lenient and disable), the default is false.

:wrench: Please make sure to familiarize yourself with [Development section of README.md](https://github.com/giantswarm/architect-orb/blob/master/README.md#development) before releasing a PR to this repository.

:warning: If you want to **create a new release** of this orb please read [Releases section of README.md](https://github.com/giantswarm/architect-orb/blob/master/README.md#releases) carefully :exclamation:

## Checklist

### DIFF
--- a/src/commands/helm-lint.yaml
+++ b/src/commands/helm-lint.yaml
@@ -1,11 +1,11 @@
 ---
 parameters:
-  lenient:
-    description: "Use more lenient lint rules"
-    type: "boolean"
-    default: false
   disable:
     description: "Disable linting alogether"
+    type: "boolean"
+    default: false
+  lenient:
+    description: "Use more lenient lint rules"
     type: "boolean"
     default: false
 description: Lint Helm charts
--- a/src/commands/helm-lint.yaml
+++ b/src/commands/helm-lint.yaml
@@ -1,5 +1,25 @@
 ---
+parameters:
+  lenient:
+    description: "Use more lenient lint rules"
+    type: "boolean"
+    default: false
+  disable:
+    description: "Disable linting alogether"
+    type: "boolean"
+    default: false
 description: Lint Helm charts
 steps:
-  - run: |
-      ct lint --validate-maintainers=false --check-version-increment=false --chart-dirs helm
+  - unless:
+      condition: <<parameters.disable>>
+      steps:
+        - when:
+            condition: <<parameters.lenient>>
+            steps:
+              - run: |
+                  ct lint --validate-maintainers=false --check-version-increment=false --validate-chart-schema=false --chart-dirs helm
+        - unless:
+            condition: <<parameters.lenient>>
+            steps:
+              - run: |
+                  ct lint --validate-maintainers=false --check-version-increment=false --chart-dirs helm
--- a/src/jobs/push-to-app-catalog.yaml
+++ b/src/jobs/push-to-app-catalog.yaml
@@ -13,6 +13,14 @@ parameters:
   chart:
     description: "Name of the chart inside helm directory to push to the App Catalog."
     type: "string"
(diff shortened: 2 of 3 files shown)


yasn77:
@stone-z This should help with https://github.com/giantswarm/ignition-operator/pull/37

stone-z:
Thanks!

src/jobs/push-to-app-catalog.yaml
```diff
@@ -13,6 +13,14 @@ parameters:
...
     description: "Name of the chart inside helm directory to push to the App Catalog."
     type: "string"
+  lenient-lint:
+    description: "Use more lenient lint rules"
+    type: "boolean"
+    default: false
+  disable-lint:
+    description: "Disable linting alogether"
+    type: "boolean"
+    default: false
```
rossf7: Nitpicky but we usual use alpha order for params unless there is a good reason not to. WDYT?

yasn77: I agree, ordering is a small thing but cognitively helps a lot. Fixed in b0237c3

src/commands/helm-lint.yaml
```diff
@@ -1,5 +1,25 @@
...
+  lenient:
+    description: "Use more lenient lint rules"
+    type: "boolean"
+    default: false
 description: Lint Helm charts
 steps:
-  - run: |
-      ct lint --validate-maintainers=false --check-version-increment=false --chart-dirs helm
+  - unless:
+      condition: <<parameters.disable>>
```
kopiczko: How about actually forcing us to add those dummy values? Would that hurt? Where could we keep them?

kopiczko: BTW I don't think it's going to work because condition works with a string where any non-empty string is true. At least that what I found in the docs back then. But if it works then great.

yasn77: Dummy values are kept in the code repo with the Helm chart (in `ci` directory`) https://github.com/helm/chart-testing/blob/master/doc/ct_lint.md

> Charts may have multiple custom values files matching the glob pattern '*-values.yaml' in a directory named 'ci' in the root of the chart's directory. The chart is linted for each of these files. If no custom values file is present, the chart is linted with defaults.

So in the long run, I expect people to add dummy values files but I didn't want to enforce that straight away as it would add additional unplanned work for people.

yasn77: Yeh, I think your right... Just saw boolean type and just presumed it would work. Just double-checked the docs it references strings only.

Will fix

src/commands/helm-lint.yaml
```diff
@@ -1,5 +1,25 @@
...
+      steps:
+        - when:
+            condition: <<parameters.lenient>>
+            steps:
+              - run: |
+                  ct lint --validate-maintainers=false --check-version-increment=false --validate-chart-schema=false --chart-dirs helm
```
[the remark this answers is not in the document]
yasn77: Ok cool... I'll look at this

src/commands/helm-lint.yaml
kopiczko: Would it mean for example `helm/ignition-operator/ci/installation-values.yaml`? So we don't even need to change the job for that right?

yasn77:
Closing and will reopen a simpler version

PR #100, opened 2020-03-11
Use `charts` instead of `chart-dirs`

yasn77:
Using `charts` implies that version bump check is disabled, it is also better to
be precise as not all charts in a directory maybe templated correctly. For
example, `push-to-app-catalog` job only takes a single chart and will template
that.

The command can still support multiple charts if need be, but now we can
also specify a single chart.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix GOPATH problem in `integration-test` job.
 - Disable version bump check for Helm lint
+- Helm lint supports single chart rather than entire directory
 
 ## [0.8.1] - 2020-03-09
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix GOPATH problem in `integration-test` job.
 - Disable version bump check for Helm lint
-- Helm lint supports single chart rather than entire directory
+- Helm lint supports single chart rather than entire directory disabling version bump check.
 
 ## [0.8.1] - 2020-03-09
 
--- a/src/commands/helm-lint.yaml
+++ b/src/commands/helm-lint.yaml
@@ -6,4 +6,4 @@ parameters:
 description: Lint Helm charts
 steps:
   - run: |
-      ct lint --validate-maintainers=false --charts <<parameters.charts>>
+      ct lint --validate-maintainers=false --charts="helm/<<parameters.charts>>"
--- a/src/commands/helm-lint.yaml
+++ b/src/commands/helm-lint.yaml
@@ -2,7 +2,7 @@
 parameters:
   chart:
     type: "string"
-description: Lint Helm charts
+description: Lint Helm chart
 steps:
   - run: |
-      ct lint --validate-maintainers=false --charts="helm/<<parameters.charts>>"
+      ct lint --validate-maintainers=false --charts="helm/<<parameters.chart>>"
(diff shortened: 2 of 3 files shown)


yasn77:
@kopiczko As discussed, simpler change.

The job passes a single chart, but the command can support comma separated list

src/commands/helm-lint.yaml
```diff
@@ -1,5 +1,9 @@
 ---
+parameters:
+  charts:
+    description: "Comma seperated list of charts to lint (relative path)"
+    type: "string"
 description: Lint Helm charts
 steps:
   - run: |
-      ct lint --validate-maintainers=false --check-version-increment=false --chart-dirs helm
+      ct lint --validate-maintainers=false --charts <<parameters.charts>>
```
kopiczko:
```suggestion
      ct lint --validate-maintainers=false --charts="helm/<<parameters.charts>>"
```

src/jobs/push-to-app-catalog.yaml
```diff
@@ -24,7 +24,8 @@ steps:
   - prepare-catalogbot-git-ssh
   - helm-chart-template:
       chart: <<parameters.chart>>
-  - helm-lint
+  - helm-lint:
+      charts: "helm/<< parameters.chart >>"
```
kopiczko: Let's just pass a single chart because the command has a **way** better output when single chart is passed.

```suggestion
      chart: "<< parameters.chart >>"
```

src/commands/helm-lint.yaml
```diff
@@ -1,5 +1,9 @@
 ---
+parameters:
+  charts:
+    description: "Comma seperated list of charts to lint (relative path)"
+    type: "string"
```
kopiczko:
```suggestion
  chart:
    type: "string"
```

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix GOPATH problem in `integration-test` job.
 - Disable version bump check for Helm lint
+- Helm lint supports single chart rather than entire directory
```
kopiczko:
```suggestion
- Helm lint supports single chart rather than entire directory disabling version bump check.
```

src/commands/helm-lint.yaml
yasn77: This will hard fix to a single chart. I would like the command to be able to support multiple charts. Would make reuse of the command better.

Not sure there is much value is forcing only a single chart, when supporting multiple charts is trivial.

src/jobs/push-to-app-catalog.yaml
yasn77: As above... Most cases like from this job, it will be a single chart so the output will be fine. Just not sure there is any value is restricting the command

src/commands/helm-lint.yaml
kopiczko: > Let's just pass a single chart because the command has a **way** better output when single chart is passed.

The value 🔼 

The output for linting schema is useless when specifying multiple charts. Also I wouldn't worry about re-usability here. It's a command not a job.

kopiczko: I checked that and I'm wrong. Apparently the output is different when `--charts-dir` is specified. Okay let's keep it if it's important to you.

src/commands/helm-lint.yaml
```diff
@@ -1,5 +1,8 @@
 ---
+parameters:
+  chart:
+    type: "string"
 description: Lint Helm charts
 steps:
   - run: |
-      ct lint --validate-maintainers=false --check-version-increment=false --chart-dirs helm
+      ct lint --validate-maintainers=false --charts="helm/<<parameters.charts>>"
```
kopiczko:
```suggestion
      ct lint --validate-maintainers=false --charts="helm/<< parameters.chart >>"
```

PR #126, opened 2020-04-09
Kat test

piontec:
## Checklist


src/commands/kat-tests-run.yaml
```diff
@@ -0,0 +1,25 @@
...
+  - run:
+      name: "start kube-app-testing tool"
+      command: |
+        PATH="/tmp/bin:$PATH" /tmp/kube-app-testing.sh -c << parameters.chart >>
+  - run:
+      name: ""
+      command: |
+        if [[ -d /tmp/kat ]]; then sudo chown -R circleci.circleci /tmp/kat; fi
+  - save_cache:
+        key: v2-python-cache-{{ checksum "test/kat/Pipfile.lock" }}
```
MarcelMue, line 18: Why is it necessary to save the cache here?

src/commands/kat-tests-install-tools.yaml
```diff
@@ -0,0 +1,40 @@
+parameters:
+  kind-version:
```
MarcelMue, line 2: How are these exposed? Can I change them through these parameters? I can't see that in the example.

src/commands/kat-tests-run.yaml
piontec: Test executor might fetch additional pip packages during run, so you have to cache them after the run.

src/commands/kat-tests-install-tools.yaml
piontec: These are not user-facing parameters (at least yet), rather templating vars to make version changes easier in the future (and potentially configurable - not really supported right now).

src/commands/kat-tests-run.yaml
MarcelMue: I don't see why we need to cache them at all - it seems like premature optimization to me to cache something when we are not sure if it actually improves build times.

src/commands/kat-tests-install-tools.yaml
MarcelMue: Okay - we should somehow make sure that we remember these / document these. Otherwise I see us diverging in versions here which is not something we want.

src/commands/helm-lint.yaml
```diff
@@ -1,8 +0,0 @@
----
-parameters:
```
MarcelMue, line 2: Do we use this command somewhere in our projects currently? If so it would be good to have a list of repos where we use it and make sure we transition them. Otherwise someone will run into this when trying to update the orb.

src/commands/kat-tests-run.yaml
piontec: I checked that before implementig caching, gives nice speed boost.

src/commands/kat-tests-install-tools.yaml
piontec: Do we have a "source fo truth" for versions we should use currently? Is it possible to get it automatically?

src/commands/helm-lint.yaml
piontec: I know just about one, kong-app, but will announce it anyway

src/commands/kat-tests-install-tools.yaml
MarcelMue: Not that I am aware of right now - something to consider for @giantswarm/wg-integration-testing  as we will use kind and similar tools there as well.

docs/run_kat_tests.md
```diff
@@ -0,0 +1,35 @@
+
+# run-kat-tests
+
+This job installs all the required software to run
+[kube-app-testing](https://github.com/giantswarm/kube-app-testing).
```
rossf7, line 5: It took me a bit to realise the kind cluster was created by the script and not here.

Maybe some more info on the split between the job and the test script would help?

src/commands/kat-tests-install-tools.yaml
rossf7: Yes having this aligned between the integration test job and here would be nice. There the kind version is hardcoded but the kubernetes version is a param.

I'm fine with this approach but yes I think something for the WG. Maybe a standard location for all the versions (kind, kubernetes, helm) would help and avoid them drifting. I'll put a note in the board.

piontec: Yeap, taht would be cool - to have a source of versions that should be used. Please ping me in the new issue as well, I would like to integrate this later with this tool.

src/commands/kat-tests-install-tools.yaml
```diff
@@ -0,0 +1,40 @@
...
+  - run:
+      name: "create temporary bin dir /tmp/bin"
+      command: |
+        mkdir /tmp/bin || true
+  - restore_cache:
+      key: bin-{{ checksum "/tmp/kube-app-testing.sh" }}
+  - run:
+      name: "install kind"
+      command: |
+        if [[ ! -f /tmp/bin/kind ]]; then curl -Lo /tmp/bin/kind "https://github.com/kubernetes-sigs/kind/releases/download/v<< parameters.kind-version >>/kind-$(uname)-amd64" && chmod +x /tmp/bin/kind; fi
```
kopiczko, line 28: It's ugly that we install kind in different places in different way. I think that should be shared by integration test and this.

fiunchinho:
Not sure I agree with moving out helm linting from `push-to-app-catalog`. I see value in linting always, even though we don't always run the `kat` test. What do we get only linting if tests are being executed? What do y'all think?

kopiczko:
> Not sure I agree with moving out helm linting from `push-to-app-catalog`. I see value in linting always, even though we don't always run the `kat` test. What do we get only linting if tests are being executed? What do y'all think?

I missed that. I also don't like that. We should lint always.

src/commands/kat-tests-install-tools.yaml
rossf7: I created https://github.com/giantswarm/giantswarm/issues/10059 for this in the WG board.

Both jobs have install tools commands so yes I think we could align them more. We use go to install in the integration test job but I'm fine with installing from a release.

fiunchinho:
Can you add it back @piontec ? Any one against it?

PR #140, opened 2020-05-07
commands/push-to-app-collection: retry the push

mig4:
Towards https://github.com/giantswarm/giantswarm/issues/10423.

Retry the `git push` up to 4 times with some wait times in between in
case it fails, in order to cope with transient errors better.

## Checklist

### DIFF
--- a/src/commands/push-to-app-collection.yaml
+++ b/src/commands/push-to-app-collection.yaml
@@ -61,9 +61,14 @@ steps:
         while ! git push; do
           attempt+=1
           if ((attempt >= 4)); then
-            echo "giving up after $attempt failures"
+            printf '%s\n' \
+              "Error pushing app to the collection. See known errors in:" \
+              "https://github.com/giantswarm/architect-orb/blob/master/README.md#push-to-app-collection" \
+              "Giving up after $attempt failures." \
+              >&2
             exit 1
           fi
+          git pull --rebase --autostash
           sleep 5
         done
   - run: |
--- a/src/commands/push-to-app-collection.yaml
+++ b/src/commands/push-to-app-collection.yaml
@@ -68,8 +68,8 @@ steps:
               >&2
             exit 1
           fi
-          git pull --rebase --autostash
           sleep 5
+          git pull --rebase --autostash
         done
   - run: |
         rm .cr_name .version .app_file_name .git_commit_message
--- a/src/commands/push-to-app-collection.yaml
+++ b/src/commands/push-to-app-collection.yaml
@@ -56,7 +56,16 @@ steps:
   - run: |
         cd .app-collection && git commit -m "$(cat ../.git_commit_message)"
   - run: |
-        cd .app-collection && git push || (echo "Error pushing app to the collection. See known errors in https://github.com/giantswarm/architect-orb/blob/master/README.md#push-to-app-collection" && exit 1)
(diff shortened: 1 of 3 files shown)


src/commands/push-to-app-collection.yaml
```diff
@@ -56,7 +56,16 @@ steps:
   - run: |
         cd .app-collection && git commit -m "$(cat ../.git_commit_message)"
   - run: |
-        cd .app-collection && git push || (echo "Error pushing app to the collection. See known errors in https://github.com/giantswarm/architect-orb/blob/master/README.md#push-to-app-collection" && exit 1)
+        cd .app-collection && \
+        declare -i attempt=0 && \
+        while ! git push; do
```
xh3b4sd, line 61: I think when you push and fail and push again it will continue to fail. We need to pull and push and when it fails try again to pull and push. Further I found `Error pushing app to the collection.` good to see so let's maybe keep that bit.

mig4:
A branch in `aws-operator` for testing this PR: https://github.com/giantswarm/aws-operator/pull/2398

src/commands/push-to-app-collection.yaml
mig4: Ah, OK, I'll work that in. Thanks.

mig4: Hey @xh3b4sd, should we keep the link to README and the explanation about re-trying manually?

mig4: PTAL

src/commands/push-to-app-collection.yaml
```diff
@@ -56,7 +56,21 @@ steps:
...
+          if ((attempt >= 4)); then
+            printf '%s\n' \
+              "Error pushing app to the collection. See known errors in:" \
+              "https://github.com/giantswarm/architect-orb/blob/master/README.md#push-to-app-collection" \
+              "Giving up after $attempt failures." \
+              >&2
+            exit 1
+          fi
+          git pull --rebase --autostash
+          sleep 5
```
xh3b4sd, line 72: What this does now is to pull and wait 5 seconds before pushing. This means we give 5 seconds more to screw it again. Let's flip it so that we wait 5 seconds and then pull and push immediately. This reduces the likelihood of getting screwed again.

mig4: Good point.

mig4: Done

mig4:
> Ugly as hell but I don't see a better way to do this now.

Best looking bash I've ever seen 😜 

But really, yeah, I considered injecting a binary that would help with that into the architect image (like [loop](https://github.com/Miserlou/Loop) or [GNU parallel](https://www.gnu.org/software/parallel/) or even something like [git-retry](https://commondatastorage.googleapis.com/chrome-infra-docs/flat/depot_tools/docs/html/git-retry.html) from chromium builds) but I suppose it would have more sense if it was used in more places, so for now I decided to give this approach a try.

mig4:
> Fine when tested

I tested the logic in a local shell, that works OK, but I'm not able to run a full job in circle because as mentioned above it looks like this job is only triggered on new releases. Unless I missed something and there's another way?

mig4:
Going with this, I'll deal with follow-ups if they're needed.

PR #143, opened 2020-05-12
Support Chart testing config file

yasn77:
This is to support the use of additional Helm chart repositories.
Config file is needed for this option according to the docs
https://github.com/helm/chart-testing/tree/v2.4.1#using-private-chart-repositories

Additional repos are needed to satisfy Chart dependencies when testing. Maybe in
the future we can copy all dependencies to our own catalogs, then we won't need
to rely on external deps and we can just add our catalogs. But until that is
in place, we need a way to specify repos.

## Checklist

### DIFF
--- a/docs/run_kat_tests.md
+++ b/docs/run_kat_tests.md
@@ -24,9 +24,8 @@ Parameters:
   and the only one supported right now.
 - `ct_config` - Path to configuration file for Chart Testing (`ct`). This is
   required, if the app relies on external dependencies. You would use this
-  config file to add additional Helm repositories. See
-  https://github.com/helm/chart-testing/tree/v2.4.1#using-private-chart-repositories.
-  Otherwise this parameter is optional
+  config file [to add additional Helm repositories](https://github.com/helm/chart-testing/tree/v2.4.1#using-private-chart-repositories).
+  Otherwise this parameter is optional.
 
 Example usage
 
--- a/docs/run_kat_tests.md
+++ b/docs/run_kat_tests.md
@@ -1,4 +1,3 @@
-
 # run-kat-tests
 
 This job installs all the required software to run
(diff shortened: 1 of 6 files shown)


src/commands/helm-lint.yaml
```diff
@@ -2,7 +2,11 @@
...
     type: "string"
+  ct_config:
+    description: Chart Testing Config file path
+    type: "string"
+    default: ""
 description: Lint Helm chart
 steps:
   - run: |
-      ct lint --validate-maintainers=false --charts="helm/<<parameters.chart>>"
+      ct lint <<# parameters.ct_config >> << parameters.ct_config >> <</ parameters.ct_config >> --validate-maintainers=false --charts="helm/<<parameters.chart>>"
```
kopiczko, line 12: Is this `#` `/` a special syntax?

yasn77: It works as a conditional for booleans -> https://circleci.com/docs/2.0/reusing-config/#boolean

Since a string is false if it's empty, I presumed it would work but it doesn't :crying_cat_face: 

I'll update to use `when`

yasn77: Actually, it does work. I was missing an CLI argument in my config.

yasn77:
@fiunchinho Doc updated, thanks for pointing it out.

docs/run_kat_tests.md
```diff
@@ -23,6 +22,11 @@ Parameters:
 - `chart` - the name of the chart to test in `/helm` directory
 - `cluster_type` - type of the cluster to create for test execution. `kind` is default
   and the only one supported right now.
+- `ct_config` - Path to configuration file for Chart Testing (`ct`). This is
+  required, if the app relies on external dependencies. You would use this
+  config file to add additional Helm repositories. See
+  https://github.com/helm/chart-testing/tree/v2.4.1#using-private-chart-repositories.
+  Otherwise this parameter is optional
```
fiunchinho, line 29:
```suggestion
  config file [to add additional Helm repositories](https://github.com/helm/chart-testing/tree/v2.4.1#using-private-chart-repositories).
  Otherwise this parameter is optional
```

PR #153, opened 2020-05-21
Add optional customer tag

pipo02mix:
## Checklist

### DIFF
--- a/docs/job/push-to-docker.md
+++ b/docs/job/push-to-docker.md
@@ -5,7 +5,7 @@ It requires the build context and Dockefile to be present at the root of worksap
 
 **NOTE**: docker registry username and password are read from environement variables which default to `ARCHITECT_DOCKER_REGISTRY_USERNAME` and `ARCHITECT_DOCKER_REGISTRY_PASSWORD` respectively. This can be changed via `username_var` and `password_var` arguments.
 **NOTE**: The docker image will be tagged with the version found by `architect project version` command.
-**NOTE**: In case you want to set a custom tag for your image define `custom-tag` with the tag value desired.
+**NOTE**: In case you want to set a custom tag for your image define `custom_tag` with the tag value desired.
 
 Example usage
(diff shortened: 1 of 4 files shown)


docs/job/push-to-docker.md
```diff
@@ -18,6 +19,7 @@ workflows:
     jobs:
       - architect/push-to-docker:
           context: "architect"
+          custom_tag: "1.0.0"
```
kopiczko, line 22: This doesn't sound like a good idea to me. We'll need to change build configuration each time we want to change the value of the tag.

pipo02mix: yeah I am working in a recipe to install gpu drivers. We had a trouble using sha tags and I would like to do it more deterministic (using driver version as tag value, so customer can set driver version in chat values and it translates easy to container image). But at the same time I want to do it using automation and not building locally. Any sideshow I can do it?

pipo02mix: I guess this will be something rare but in certain scenarios can be useful

pipo02mix:
It works https://circleci.com/gh/giantswarm/kubernetes-gpu/26

docs/job/push-to-docker.md
kopiczko: Why then not releasing the repo with the gpu dirvers and having nice tag?

kopiczko: You can `docker pull quay.io/giantswarm/aws-operator:8.6.0`

pipo02mix: well I guess it can be an option. I was thinking it is more intuitive if the driver container image is tagged with the driver version it contains (so the customer does need to map versions) like `coreos-nvidia-driver-installer:390.116`

pipo02mix:
Ok Pawel and I discussed. Using the release tag functionality it is enough. Closing

PR #216, opened 2020-12-02
Update steps names to align and make them grep-friendly

kopiczko:


TheoBrigitte:
Looks fine but I do not get the use case for this.

kopiczko:
It solves two issues for me:

1. By looking at the job output I can tell from which command this step comes from (circle glues all the configs together so it's difficult to tell otherwise)
2. I can grep for `architect/push-to-app-collection: Generate app CR` and find the step immediately. When there is some string interpolation this requires some reverse engineering.

kopiczko:
Check below, you can tell what command/step was executed:

https://app.circleci.com/pipelines/github/giantswarm/config-controller/255/workflows/9e03f56c-1ab1-4983-9679-3f90767a9d69/jobs/849

Screenshot 2020-12-03 at 10 25 35

TheoBrigitte (approved):
Ok makes sense

PR #295, opened 2021-07-29
Split off app testing

ubergesundheit:
This PR contains changes in regards to the app-build-suite/app-test-suite split.

It completely removes the `run-tests-with-abs` job and replaces it with `run-tests-with-ats`. This is a breaking change.

Edit: We decided this breaking is a good thing. Changing up repositories is straight forward. Examples: https://github.com/giantswarm/cert-manager-app/pull/179 https://github.com/giantswarm/nginx-ingress-controller-app/pull/223 https://github.com/giantswarm/efk-stack-app/pull/93

In regards for easing the transition we have multiple options:

- Keep the `run-tests-with-abs` job for now, to ease transition to `run-tests-with-ats` (my favourite).
  - Problem here is, that the updated `app-build-suite` does not like the testing options in the `.abs/main.yaml` file, so I think this is going to crash about 100%. Maybe @piontec we can release a version not crashing for unknown configuration options?
- Write a converter script to help making the required changes to app repositories (rename jobs and options in circle config, move test files)
- Just do nothing and let CI runs crash

## Checklist


docs/job/run-tests-with-ats.md
```diff
@@ -0,0 +1,73 @@
...
+
+Actual implementation is:
+
+```
+find -name "<< parameters.chart_archive_prefix >>*.tgz" -print -quit
+```
+
+### app-test-suite_version
+
+Version of app-test-suite `dabs.sh` container wrapper to use (git tag or commit).
```
piontec, line 54: Should it be `dats.sh` here?

docs/job/run-tests-with-ats.md
```diff
@@ -0,0 +1,73 @@
...
+Use this parameter if you have some changes lined up in `dabs.sh` which is not released yet.
```
piontec, line 55: ... and here?

docs/job/run-tests-with-ats.md
```diff
@@ -0,0 +1,73 @@
...
+
+Container tag of app-test-suite to use (check [quay.io/giantswarm/app-test-suite](https://quay.io/giantswarm/app-test-suite)).
+This parameter allows to specify the used container tag of app-test-suite.
+
+(Default: "0.1.2")
+
+### additional_app-test-suite_flags
+
+Allows to add additional flags to the execution of app-test-suite.
+If possible, specify your configuration using the `.abs/main.yaml` configuration file.
```
piontec, line 73:
```suggestion
If possible, specify your configuration using the `.ats/main.yaml` configuration file.
```

piontec (approved):
LGTM. One thought though: we should use circleci caching more extensively. Both `pipenv` and `pre-commit` use caching extensively and make builds much faster.
Here's the location for `pre-commit` caching: https://github.com/giantswarm/step-exec-lib/blob/b5ceb8c55b817a238274fd341dc1b7beb48ad1b2/.github/workflows/main.yml#L62
`pipenv`'s in `~/.cache/pipenv`.
AFAIR this made my builds go down from like 5m to 2m. Would be super cool to include this caching in architect-orb.
Actually, that's a separate ticket, but came to my mind just now :P

ubergesundheit:
Thanks, fixed the copy & paste artifacts in the docs

Regarding caching: That would be very awesome! At this point the `dats.sh` wrapper is called and everything is happening inside the container afterwards. So we would need to mount caches to the container in some place.

Do you know where the cache would need to be mounted?

Another way of doing this would be to install everything "by hand" (python, dependencies, kind binary, ...) then cache these files.

I think it would be beneficial to bundle a pre-warm cache into the container image of app-test-suite. Do you think this would be possible?

piontec:
Caches are written in the container, I gave their locations above. We have to map them somewhere on the disk of the host, then save to circleci cache.
As for including in the image... Yes, that's rather easy to do, but these deps change quite frequently, especially with renovate keeping track of updates... it might work, might be too big to include in the image. Let's try :) 
But my first step would be around circle ci's caching. If you want, we can try to do it together in our working hours.

ubergesundheit:
For upgrade testing, we identified the requirement of the `run-tests-with-ats` job to push its test artifacts to the catalogs.

> So, the CI workflow as I see it goes like this:
> - CI checks out catalog repo
> - CI invokes `ats`
> - `ats` produces 1 new file in `my-app-0.1.2.tgz-meta/tested-upgrade-0.1.2.3.yaml`
> - CI does commit & push on the repo

[Source](https://github.com/giantswarm/giantswarm/pull/18822#discussion_r703457659)

piontec:
As for the unknown config options, I think it should be either the 'crash' or 'ignore unknown' solution. I think I actually like 'crash' better, as otherwise, we're creating confusion (or impression) that some tests are being run, as in reality they are not. To be short, I like explicit better, but that's not a very strong opinion. WDYT?

ubergesundheit:
Yeah making builds break is a good way to make people aware of change. So lets leave it like this and break peoples builds ;)

src/commands/run-tests-with-ats.yaml
```diff
@@ -0,0 +1,35 @@
...
+      command: |
+        export DATS_TAG=<< parameters.app-test-suite_container_tag >>
+        # find the chart archive file name
+        chart_archive=$(find build -name "<< parameters.chart_archive_prefix >>*.tgz" -print -quit)
+        # move chart archive from build directory to root
+        # we're doing this to make sure .ats/main.yaml is picked up
+        cp "$chart_archive" "${chart_archive##*/}"
+
+        # execute dats wrapper
+        ./dats.sh --chart-file "${chart_archive##*/}" << parameters.additional_app-test-suite_flags >>
```
MarcelMue, line 35: Slightly off topic but something I have been wondering:
Can we reduce the permissions / settings required for `dats.sh`? AFAIK it will still not work like this with e.g. podman.

docs/job/run-tests-with-ats.md
```diff
@@ -0,0 +1,73 @@
...
+
+Container tag of app-test-suite to use (check [quay.io/giantswarm/app-test-suite](https://quay.io/giantswarm/app-test-suite)).
+This parameter allows to specify the used container tag of app-test-suite.
+
+(Default: "0.1.2")
+
+### additional_app-test-suite_flags
+
+Allows to add additional flags to the execution of app-test-suite.
+If possible, specify your configuration using the `.ats/main.yaml` configuration file.
```
MarcelMue, line 73: Would be cool to have a little sample file for `.ats.main.yaml` or ling to an example.

src/commands/run-tests-with-ats.yaml
ubergesundheit: Created https://github.com/giantswarm/giantswarm/issues/18884

Feel free to add your observations in the issue!

PR #309, opened 2021-09-29
Change `kubeval` k8s schema to more up-to-date source

TheoBrigitte:
## Checklist

### DIFF
--- a/src/commands/kubeval.yaml
+++ b/src/commands/kubeval.yaml
@@ -11,6 +11,7 @@ steps:
         KUBERNETES_VERSIONS: 1.16.0 1.17.0 1.18.0 1.19.0 1.20.0
       command: |
         for v in $KUBERNETES_VERSIONS; do
+          echo "Validating against kubernetes version $v"
           if [ -d "helm/<<parameters.chart>>/ci" ]; then
             for t in $(ls helm/<<parameters.chart>>/ci/*.yaml); do
               helm template --values $t helm/<<parameters.chart>> | kubeval --ignore-missing-schemas --kubernetes-version $v -
--- a/src/commands/kubeval.yaml
+++ b/src/commands/kubeval.yaml
@@ -8,9 +8,14 @@ steps:
       name: kubeval
       environment:
         KUBEVAL_SCHEMA_LOCATION: https://raw.githubusercontent.com/yannh/kubernetes-json-schema/411622f5e91bc54d3eff7b700661ba03f23f7729
+        KUBERNETES_VERSIONS: 1.16.0 1.17.0 1.18.0 1.19.0 1.20.0
       command: |
-        if [ -d "helm/<<parameters.chart>>/ci" ]; then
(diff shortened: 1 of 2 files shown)


src/commands/kubeval.yaml
```diff
@@ -8,7 +8,7 @@ steps:
       name: kubeval
       command: |
         if [ -d "helm/<<parameters.chart>>/ci" ]; then
-          for t in $(ls helm/<<parameters.chart>>/ci/*.yaml); do helm template --values $t helm/<<parameters.chart>> | kubeval --ignore-missing-schemas - ; done
+          for t in $(ls helm/<<parameters.chart>>/ci/*.yaml); do helm template --values $t helm/<<parameters.chart>> | kubeval --ignore-missing-schemas --kubernetes-version skip - ; done
```
fiunchinho, line 11: question: without this change, would `kubeval` do the same as `deprek8ion`? What I'm asking is, could we remove `deprek8ion` and use only `kubeval` or are they different things?

TheoBrigitte: Yes, without this change, would `kubeval` do the same as `deprek8ion`. We could in theory remove `deprek8ion` and use only `kubeval`.

I introduce `kubeval` because it can detect overlapping k8s resources as well.

fiunchinho: Cool. No blocking here, but we should definitely take a look and see if we can get rid of `deprek8ion` and make things simpler.

fiunchinho (approved):

src/commands/kubeval.yaml
TheoBrigitte: How about we leave this here, see if we are happy with `kubeval` schema validation and drop `deprek8ion` later on then ?

TheoBrigitte:
Update kubernetes schema location

from
-  https://github.com/instrumenta/kubernetes-json-schema (default) last updated on Apr 29, 2020, up to `v1.18.1`

to
- https://github.com/yannh/kubernetes-json-schema last updated on Sep 25, 2021, up to `v1.22.2`

src/commands/kubeval.yaml
```diff
@@ -6,9 +6,11 @@ description: Validate Kubernetes resources.
 steps:
   - run:
       name: kubeval
+      environment:
+        KUBEVAL_SCHEMA_LOCATION: https://raw.githubusercontent.com/yannh/kubernetes-json-schema/master
```
ubergesundheit, line 10: is pointing at `master` a good idea? What if it changes and suddenly all builds break because of a change we don't know of?

TheoBrigitte: Good point, but this repository has 1 branch, 0 tags, the best I can do is point to current master commit `411622f5e91bc54d3eff7b700661ba03f23f7729`. Does this works for you?

ubergesundheit (approved):

TheoBrigitte:
Going with this
image

PR #335, opened 2022-02-10
Use docker buildkit

ubergesundheit:
This PR switches `push-to-docker` job to use [Docker Buildkit](https://docs.docker.com/develop/develop-images/build_enhancements/) and docker 20.10.11


fiunchinho (approved):
nice!

PR #362, opened 2022-04-19
Add a command to push to OCI registries

kubasobon:
Towards https://github.com/giantswarm/roadmap/issues/391

## Checklist

### DIFF
--- a/src/commands/package-and-push-oci.yaml
+++ b/src/commands/package-and-push-oci.yaml
@@ -59,7 +59,7 @@ steps:
   - run:
       name: Package the chart archive
       command: |
-        mkdir build && helm package ./helm/<< parameters.chart >> --destination ./build
+        mkdir -p build && helm package ./helm/<< parameters.chart >> --destination ./build
   - when:
       condition: << parameters.persist_chart_archive >>
       steps:
--- a/src/commands/package-and-push-oci.yaml
+++ b/src/commands/package-and-push-oci.yaml
@@ -1,0 +1,94 @@
+parameters:
+  app_catalog:
+    type: "string"
+    description: |
+      Set this to an OCI registry with a specific subpath, e.g.
+      `oci://giantswarmpublic.azurecr.io/charts/`. Last slash is mandatory.
+  app_catalog_test:
+    type: "string"
+    description: |
+      Set this to an OCI registry with a specific subpath, e.g.
+      `oci://giantswarmpublic.azurecr.io/test_charts/`. Last slash is mandatory.
+  chart:
+    type: "string"
+  on_tag:
+    type: boolean
+    default: true
+    description: |
+      When this is `false`, commits to `master` will be pushed to `app_catalog` instead of `app_catalog_test`.
+      Set this to `false` for deployments that follow a a master branch for production releases rather than
+      using tags (the default).
+  explicit_allow_chart_name_mismatch:
+    type: boolean
+    default: false
+    description: |
+      If 'explicit_allow_chart_name_mismatch' is set to true, the name of the chart can be anything.
+      Otherwise the name set in the 'chart' parameter must start with the repository name and optionally continue with '-app'.
+      Does not have any effect for 'executor: app-build-suite'.
+  persist_chart_archive:
+    type: boolean
+    default: false
+    description: |
+      When this is `true`, the packaged chart archive will be persisted to the workspace.
+      Set this to `true`, if you're planning to execute tests using app-test-suite.
+steps:
+  - when:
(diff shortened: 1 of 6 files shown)


src/commands/package-and-push-oci.yaml
```diff
@@ -0,0 +1,94 @@
+parameters:
+  app_catalog:
```
MarcelMue, line 2: Should we set defaults on these?

kubasobon: I like this idea a lot. Done.

src/commands/package-and-push-oci.yaml
```diff
@@ -0,0 +1,96 @@
...
+      command: |
+          readonly app_catalog_name="$(cat .app_catalog_name)"
+          readonly reference="$(cat .reference)"
+
+          ret=1
+          tries=4
+          for i in $(seq 1 $tries) ; do
+            echo "====> Attempt $i: Running: helm push ../build/*.tgz $app_catalog_name"
+            set +e
+            helm push ../build/*.tgz $app_catalog_name
```
MarcelMue, line 83: Maybe a stupid question: Where does helm actually come from here? Is it just installed by default? And should we verify that the version is recent enough?

MarcelMue (approved):
Generally looking good to me - just a small question :)

src/commands/package-and-push-oci.yaml
```diff
@@ -0,0 +1,96 @@
...
+        mkdir build && helm package ./helm/<< parameters.chart >> --destination ./build
+  - when:
+      condition: << parameters.persist_chart_archive >>
+      steps:
+        - persist_to_workspace:
+            root: build
+            paths:
+              - "<< parameters.chart >>*.tgz"
+  - run:
+      name: Push chart archive to OCI registy app catalog
```
rossf7, line 73:
```suggestion
      name: Push chart archive to OCI registry app catalog
```

rossf7 (approved):
LGTM

src/commands/package-and-push-oci.yaml
kubasobon: `architect-orb` uses `architect`'s Dockerfile as a base image. We install a bunch of binaries in that Dockerfile, including helm. I've recently updated `architect` (https://github.com/giantswarm/architect/pull/728/files), and `architect-orb` is already using the latest release.

kubasobon:
> Thats good! TBH a solution where the git based pushing would be moved into a separate command and then making the pushing command/mechanism configurable in the existing jobs would be less confusing, but this is also fine for me.

Bundling the two sounds like a solid idea. I'll add it to the proposed PR as well.

MarcelMue (approved):
It still makes sense to me - I struggle a bit to understand how this looks like in a project in the end, but that's a minor issue.

kubasobon:
@MarcelMue This PR adds a new command set for pushing charts to OCI registries (two new commands). It also adds two switches to the `push-to-app-catalog` job. `push_to_appcatalog`, enabled by default, runs existing commands and publishes charts to GitHub. `push_to_oci_registry`, disabled by default, runs new commands and publishes charts to the AzureCR OCI registry. 

If we merge this and craft a release, everything will remain the same due to the defaults. If we flip the switches we can change behavior to push to OCI as well or push to it exclusively.

Do we want `push_to_oci_registry` right away?

MarcelMue:
I think it makes sense to release first with the switch disabled and then try it in select repositories. WDYT?

kubasobon:
I agree :man_scientist:

kubasobon:
It works!
Screenshot_20220421_120317
Screenshot_20220421_120259

kubasobon:
@ubergesundheit PTAL. I think this covers all your requests. The only thing I had to do to enable this is add the `push_to_oci_registry: true` to my test branch.

```yaml
workflows:
  version: 2
  build-and-push:
    jobs:
      - architect/push-to-app-catalog:
          context: "architect"
          executor: "app-build-suite"
          name: "package and push chart to app catalog"
          app_catalog: "giantswarm-catalog"
          app_catalog_test: "giantswarm-test-catalog"
          chart: "hello-world-app"
          # Trigger job on git tag.
          filters:
            tags:
              only: /^v.*/
          push_to_oci_registry: true
```

ubergesundheit:
Thanks :) Some addition to the [docs](https://github.com/giantswarm/architect-orb/blob/master/docs/job/push-to-app-catalog.md) would have been nice

kubasobon:
Sure, on it.

PR #366, opened 2022-05-05
new docker process

TheoBrigitte:
Attempt at re-vamping the current docker build/push process.

Current process drawbacks:
* could end up on slightly different images being pushed to different registries
* wasted time building the same image


TheoBrigitte:
Does not seems to work 
image

`One DLC volume can only be attached to one machine or Remote Docker job at a time` https://circleci.com/docs/2.0/docker-layer-caching/#how-dlc-works

TheoBrigitte:
Honeybadger is working on something similar, leaving this out.

PR #375, opened 2022-05-31
revert non-working .nancy-ignore.local

TheoBrigitte:
Revert `.nancy-ignore.local` file, which does not work in combination with the current `.nancy-ignore` file, as `nancy` only accept one `--exclude-vulnerability-file` and take the latest.


ubergesundheit (approved):

uvegla:
😢

PR #469, opened 2023-11-27
rework dev uploads

piontec:
## Checklist

- updates to `push-to-registries`


docs/job/push-to-registries.md
```diff
@@ -25,6 +25,7 @@ workflows:
       - architect/push-to-registries:
           context: architect
           name: push-to-registries
+          image: giantswarm/IMAGE_NAME
```
marians, line 28: Since `image` it's optional now, can we please remove it here, so that this works as a minimal example?

src/jobs/push-to-registries.yaml
```diff
@@ -4,9 +4,9 @@ executor: "architect"
 resource_class: "<< parameters.resource_class >>"
 parameters:
   image:
-    description: Name of the  container repository and image. Defaults to org-name/repo-name. Must not contain registry host name!
+    description: Name of the  container repository and image. Defaults to `giantswarm/REPO_NAME`. Must not contain registry host name!
     type: string
-    default: ""
+    default: "giantswarm/${CIRCLE_PROJECT_REPONAME}"
```
marians, line 9: Have you tested this defaulting?

marians (approved):
What i see looks good. I'd like if we would remove the `image` parameter from the example, but no blocker.

PR #480, opened 2023-12-06
add better retries

piontec:
- better retries


marians (approved):
LGTM if it works for you

PR #487, opened 2024-01-08
Add job `go-dependencies` for checking a Go projects dependencies

marians:
The idea here is to call nancy in a job separate from `go-test` or `go-build` eventually, and make it non-blocking (not required).

The first step is to add a new job (this PR) which then has to be added to CircleCI configurations.

Later, the nancy execution will have to be removed from go-test and go-build.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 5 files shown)


QuentinBisson:
@marians is this still needed?

marians:
I think the intent is still valid, but I forgot about it

stone-z:
FWIW `nancy` is intended to be required and minimally blocking. Teams can suppress findings at their discretion to unblock themselves, but the scan itself fulfills compliance requirements

PR #507, opened 2024-03-05
Make prepare image tag command fail when architect call fails

uvegla:
Towards: https://github.com/giantswarm/giantswarm/issues/30055

The command substitution hides the return code from the `architect project version` call. CircleCI runs the script with `-eo pipefail` ([reference](https://circleci.com/docs/configuration-reference/#default-shell-options) + [example](https://app.circleci.com/pipelines/github/giantswarm/giantswarm/11900/workflows/50a61db0-579f-49df-a8b5-0502b8d3d35d/jobs/15865?invite=true#step-104-0_69)).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Make `image-prepare-tag` command fail when `architect project version` command fails (because CircleCI runs the scripts with `-o pipefail`)
+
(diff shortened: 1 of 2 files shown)


marians (commented):
So why would this fail now, and why did not fail before? Is it because `architect project version` was executed in an echo statement?

uvegla:
> So why would this fail now, and why did not fail before? Is it because `architect project version` was executed in an echo statement?

Not because of echo, but because of the sommand substitution: `$(...)` the exit code is masked.

marians (approved):

PR #511, opened 2024-03-13
Support resource_name and creation when generator does not exist

uvegla:
Towards: https://gigantic.slack.com/archives/C02GDJJ68Q1/p1710328832294499

## Checklist

- [ ] Make yourself familiar with following readme sections:
    - [ ] [Coding Standards](https://github.com/giantswarm/architect-orb#coding-guidelines).
    - [ ] [Development](https://github.com/giantswarm/architect-orb#development).
    - [ ] [Design and Goals](https://github.com/giantswarm/architect-orb#design-and-goals).
- [ ] :warning: Update changelog in [CHANGELOG.md](https://github.com/giantswarm/architect-orb/tree/master/CHANGELOG.md).

PR #556, opened 2024-09-19
Add `resource_name` parameter to `push-to-app-collection` job

TheoBrigitte:
This PR adds a `resource_name` parameter to the `push-to-app-collection` job. This parameter is used to specify the name of the app that is being pushed to the app collection.
The name being the final name of the app when deployed into a cluster.


QuentinBisson:
I think @uvegla already had a pr here  https://github.com/giantswarm/architect-orb/pull/511/files

PR #602, opened 2025-03-27
Add Kyverno apply job

fhielpos:
Towards: https://github.com/giantswarm/giantswarm/issues/32953


stone-z:
Not requesting changes at this point just dropping it here for your consideration:
It would be nice to not template the chart twice, and to use the chart as rendered by `abs` to take advantage of the placeholder templating and other shenanigans in some charts.
You might consider trying to call this as part of the existing [`architect`](https://github.com/giantswarm/architect-orb/blob/main/src/jobs/push-to-app-catalog.yaml#L121) and/or [`abs`](https://github.com/giantswarm/architect-orb/blob/main/src/jobs/push-to-app-catalog.yaml#L141) build flows (maybe as a [Giant Swarm validator](https://github.com/giantswarm/app-build-suite/blob/main/docs/helm-build-pipeline.md) step)

fhielpos:
Makes sense, will check that. We still need to render kyverno-policies somehow and I'm not fully sure if architect can do it in that context, but will take a look

fhielpos:
I would like this to be done during chart building, but that would just be a lot, would need to add it to abs, architect, and then architect-orb most likely.

The push-to-app-catalog step does not make a lot of sense to me since that step can be repeated for diferent catalogs and would mean the test running multiple times, which is fine but really not necessary.

stone-z:
> that step can be repeated for diferent catalogs and would mean the test running multiple times

Totally agree, it would be nice if `abs` and the orb separated the chart build from the push, or have an API for us to register checks or something. But it's currently all in one AFAICT.

fhielpos:
I will evaluate the abs route, if it makes sense and isn't too tedius I will do it there

fhielpos:
I will leave it as-is for now. Tried going the `push-to-catalogs` route but since those are different executors, `abs` should have all the binaries, which makes it fail now since it can't do curl or anything like that.

fhielpos:
This works, so I'm going to merge and release this. Improvements will come in the future 🚀

ssyno (approved):

PR #718, opened 2026-03-19
v9: Generate tags according to the new RFC

piontec:


piontec:
replaced by the `gitsemver` integration

PR #735, opened 2026-05-07
Fix multi-arch build & push regressions

QuentinBisson:
## Summary

Five surgical fixes to the v8.0.x line. No behavior changes for callers other than `os:` on `go-build` (now ignored — was already documented as deprecated).

### Fixed

- **Tag-suffix double-append in multi-arch push.** `image-prepare-tag` already bakes `tag-suffix` into `DOCKER_IMAGE_TAG`; `image-build-and-push-multiarch` then emitted a third tag `${TAG_VALUE}-${tag_suffix}` producing e.g. `1.2.3-debug--debug`. Removed the redundant block — the kept tag list now matches the single-arch path (`image-push-to-registries`).
- **`.ldflags-<GOOS>-<GOARCH>` lookup in `go-build` always missed.** `go-test` writes a single arch-agnostic `.ldflags`; the multi-arch path looked for per-arch files that nothing produces and silently fell back to `-s -w`. Result: `gitSHA` and `buildTimestamp` were stamped into native builds but **not** into cross-compiled binaries. Now reads `.ldflags`.
- **Unreachable legacy branch in `go-build`.** `architecture` defaults to `linux/amd64`, so the `[[ -n "$architecture" ]]` guard was always true and the `os`-based branch was dead code. Removed it. The `os` parameter is kept (default `""`) for backward compatibility but explicitly ignored — it stays around until v9.
- **Silent failure on GitHub repo visibility check.** Both the multi-arch and single-arch push commands ran `jq .private` on the raw `curl` body. On 403 / 429 / 5xx the response is `{"message": "..."}` and `jq .private` returns `null`, which compares unequal to `"false"` so the image was treated as **private** — silently dropping pushes to public registries. Now: capture HTTP status separately, fail with the response body when status ≠ 200 or `.private` doesn't parse to a boolean. Mirrors the v6.15.0 fix already applied to `push-helm.yaml`.
- **`setup_remote_docker: version: default`** is unspecified — drifts with whatever CircleCI defaults to today. Pinned to `docker24` in `push-to-registries` and `push-to-registries-multiarch`. Bump deliberately when a newer version is needed.

Target version: v8.0.1 (patch — no API breakage).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Drop the duplicated `<version>-<suffix>-<suffix>` tag emitted by the multi-arch push when `tag-suffix` was set (the suffix is already baked into `DOCKER_IMAGE_TAG`).
+- Read the single `.ldflags` file (written by `go-test`) in the multi-arch `go-build` path. The previous lookup of `.ldflags-<GOOS>-<GOARCH>` always missed and silently dropped `gitSHA` / `buildTimestamp` from cross-compiled binaries.
+- Remove the unreachable legacy branch in `go-build` (the `architecture` default of `linux/amd64` made the `os`-based branch dead code). The `os` parameter is kept for backward compatibility but is now ignored; use `architecture` instead.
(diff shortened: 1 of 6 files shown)


QuentinBisson:
@piontec or @uvegla do you think you could take a look?

uvegla (approved):

uvegla:
Lgtm, ty for doing this! 🙇🏻

PR #736, opened 2026-05-07
Simplify multi-arch API: architectures list, auto-derive platforms, OCI labels

QuentinBisson:
Target version: **v8.1.0**.

### Added

- `architectures` (CSV) parameter on `go-build`. Builds all targets in one job, writes `.platforms` to the workspace. Removes the need for a CircleCI matrix at the consumer.
- `push-to-registries` auto-derives `--platform` from `.platforms` when the explicit `platforms` parameter is empty.
- Standard OCI image labels (and matching index annotations in multi-arch mode): `org.opencontainers.image.{source,revision,version,created}`. `created` is taken from the commit timestamp so rebuilds of the same SHA produce identical labels; `version` is omitted when no tag is available. Opt out with `oci-labels: false`.

### Changed

- `image-login-to-registries`: `--password-stdin` / `--pass-stdin` instead of building shell command strings; bash indirect expansion instead of `eval`; read the file directly so a failed login propagates out of the loop.

### Fixed

- `image-login-to-registries`: turn off `set -x` before resolving credentials so passwords no longer leak into CI logs.

### Deprecated

- `os` parameter on `go-build` is now ignored; use `architectures` (or singular `architecture` for matrix callers).


QuentinBisson:
cc @uvegla if you have some time :)

piontec (approved):
Not tested, but the changes look really good to me!

PR #720, opened 2026-03-31
Update app-test-suite to v0.14.0

iuriaranda:
## Checklist

- [ ] Make yourself familiar with following readme sections:
    - [ ] [Coding Standards](https://github.com/giantswarm/architect-orb#coding-guidelines).
    - [ ] [Development](https://github.com/giantswarm/architect-orb#development).
    - [ ] [Design and Goals](https://github.com/giantswarm/architect-orb#design-and-goals).
- [x] :warning: Update changelog in [CHANGELOG.md](https://github.com/giantswarm/architect-orb/tree/master/CHANGELOG.md).

PR #738, opened 2026-05-07
renovate: track app-test-suite version and container tag

QuentinBisson:
## Summary

- Add Renovate `customManagers` covering the `app-test-suite_version` (git tag, prefixed `v`) and `app-test-suite_container_tag` (container tag, no prefix) defaults in `src/jobs/run-tests-with-ats.yaml`.
- Group both into a single Renovate PR via `packageRules.groupName: app-test-suite` so the git tag and container tag stay in lockstep.
- Replaces the manual bump in #720; once merged, future ATS releases land via Renovate.

## Notes

- Datasources: `github-tags` for `giantswarm/app-test-suite`, `docker` for `gsoci.azurecr.io/giantswarm/app-test-suite`.
- The two `matchStrings` are scoped to the job file so unrelated `default:` lines aren't matched.
- `app-build-suite` already has a `packageRule` and is picked up by the standard CircleCI/Docker manager via `src/executors/app-build-suite.yaml` — no change needed there.

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -3,11 +3,39 @@
     // Base config - https://github.com/giantswarm/renovate-presets/blob/main/default.json5
     "github>giantswarm/renovate-presets:default.json5",
   ],
+  "customManagers": [
+    {
+      "customType": "regex",
+      "fileMatch": ["^src/jobs/run-tests-with-ats\\.yaml$"],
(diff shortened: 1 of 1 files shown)


piontec (approved):
LGTM

PR #740, opened 2026-05-07
v8.2: Register QEMU/binfmt and fix buildx bootstrap

QuentinBisson:
Target version: **v8.2.0**. Stacked on top of #736 (v8.1 multi-arch simplification); merge that first.

### Added

- **`tonistiigi/binfmt --install all`** runs as a privileged one-shot container before the buildx build. This registers QEMU handlers in the kernel so Dockerfiles with `RUN` steps on a non-host architecture (e.g. plain `RUN apk add` or `RUN go build` without `--platform=\$BUILDPLATFORM`) build correctly under buildx multi-arch. Image tag tracked by Renovate.
- **Explicit buildx bootstrap.** The previous `docker buildx create --use --name multiarch-builder 2>/dev/null \|\| docker buildx use multiarch-builder` masked any creation failure (network, permissions, daemon issue) by silently falling back to a `use` that would also fail later, with a confusing error. Now: `inspect` first, `create --driver docker-container --use` if missing, then `inspect --bootstrap` to ensure the buildkit container is up.

### Why

Today the orb works for the dominant Giant Swarm Dockerfile patterns (`ADD ./bin-\${TARGETARCH}` or `--platform=\$BUILDPLATFORM` builders) without QEMU. But those patterns are project-by-project conventions, not contractual. Any project that writes a plain Dockerfile (`RUN apk add ca-certificates`, `RUN go build`, etc.) and bumps to v8.2 should get a working multi-arch image without having to retro-engineer arch-aware Dockerfile patterns.

### Trade-off

Emulated `RUN` is 5–20× slower than native execution. For Dockerfiles that already use `--platform=\$BUILDPLATFORM` for the heavy work (muster pattern), binfmt being registered is a no-op — nothing emulates. For plain Dockerfiles, builds work but may be slow; that's the explicit trade-off vs. forcing every project to refactor.

Image pinning: `tonistiigi/binfmt:qemu-v8.1.5` with a Renovate annotation. Reviewer: feel free to bump to a newer pinned tag, or replace with a digest pin if you prefer that supply-chain posture.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  architect: giantswarm/architect@8.0.1
+  architect: giantswarm/architect@8.0.2
   orb-tools: circleci/orb-tools@12.4.0
 
 workflows:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `image-login-to-registries`: docker and regctl logins now pipe the password via stdin (`--password-stdin` / `--pass-stdin`) instead of building a shell command string. Drops the `eval`-based env-var resolution in the regctl branch in favour of bash indirect expansion.
 - `image-login-to-registries`: read the registries data file directly (`while read … done < .registries_data`) instead of piping through `cat`, so a failed login terminates the script instead of being trapped in a subshell.
 - `go-build`: validate the `architectures` parameter against `^[a-zA-Z0-9/_,-]+$` before splitting, matching the existing check on `platforms`.
+- Update `run-tests-with-ats` job with latest app-test-suite version (v0.15.0).
 
 ### Fixed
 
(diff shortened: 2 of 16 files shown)


QuentinBisson:
Tested here https://github.com/giantswarm/observability-operator/pull/821

piontec (commented):
Hmm, forcing users to migrate to a new config by slowing them down 5-20x doesn't sound nice. Maybe let's first enable the migration, then ask users to migrate, with a reasonable deadline, and only then make the switch?

QuentinBisson:
How would you do this? Make amd64 the only architecture by default for now? Considering we're working on arm support, it might be better to enable it everywhere imo

QuentinBisson:
By the way, the delay is only for dockerfiles that do use RUN which we probably don't have a lot of considering we use go-build in most cases?

QuentinBisson:
I surveyed all GS Dockerfiles. The 5-20× slowdown only triggers when a consumer opts into multi-arch (architectures: linux/amd64,linux/arm64), which today defaults to amd64-only. binfmt registration on its own is a no-op when target arch == host arch same speed as today so projects would not be affected.

Bucket breakdown across 214 repos: 
- ~25% pre-built-binary pattern (fine)
- ~32% already use --platform=$BUILDPLATFORM (fine)
- ~36% have a naive Dockerfile and would emulate (~78 repos)
- ~7% are deliberately single-arch.

piontec:
OK, let's merge it, but let's also create a doc that explains how to convert your Dockerfile to correct non-emulated build. Then, we link this in release announcement and changelog.

QuentinBisson:
Do you mind doing that in this PR ? Or maybe we do it just before the last v8 release?

QuentinBisson:
@piontec I updated the docs in the orb and added a pointer here https://github.com/giantswarm/giantswarm/pull/36620. We probably want to revamp the doc after the v9 release?

piontec (approved):
LGTM

PR #745, opened 2026-05-07
v8.3: Provenance, SBOM, and cosign signing for multi-arch push

QuentinBisson:
Target: **v8.3.0**. Stacked on #740 (v8.2 binfmt) → #736 (v8.1 simplification).

Adds three supply-chain features on the multi-arch push path, all on by default:

- **`provenance: min|max|false`** (default `min`) — SLSA provenance attestation. `max` is opt-in only because it embeds `--build-arg` values verbatim, which would leak secrets passed that way.
- **`sbom: true|false`** (default `true`) — SPDX SBOM attestation per platform.
- **`sign: true|false`** (default `true`) — cosign keyless signing. Public images only; private skipped to avoid leaking digests/timestamps to public Rekor. Mints a fresh OIDC token with `aud=sigstore` via `circleci run oidc get` (the auto-injected `CIRCLE_OIDC_TOKEN_V2` has the wrong audience for Fulcio).

Cosign signatures land as proper OCI 1.1 referrer artifacts (clean, verifiable via the registry referrers API). Provenance + SBOM are inline because BuildKit doesn't support referrer-mode attestation storage today — confirmed against Docker docs, the `--attest` flag reference, and `buildkitd.toml` reference. Will revisit when BuildKit ships it.

`cosign` itself is now assumed on PATH (baked into the architect image via giantswarm/architect#1314); the inline `curl + chmod` install was dropped.

Verified end-to-end with the v8.3 dev orb against observability-operator (test PR linked below): manifest is multi-arch, both arch binaries are correct (`x86-64` / `ARM aarch64`), cosign signature is queryable as a referrer, and the Sigstore cert's source-repo extension (OID `1.3.6.1.4.1.57264.1.12`) populates `github.com/giantswarm/observability-operator` so verification policies can pin to readable identity rather than CircleCI's UUID-based SAN URI.

Test PR: giantswarm/observability-operator#821


piontec (commented):
Super nice! And it uses a custom OIDC token to fix the `aud` problem present since forever in circle! Have you validated the SBOM and signature after the build?

QuentinBisson:
I added a verify step later in the pipeline https://github.com/giantswarm/architect-orb/pull/749

piontec (approved):

PR #746, opened 2026-05-07
Buildx registry layer cache (cache: true|false)

QuentinBisson:
Target: **v8.4.0**. Stacked on #745 (v8.3) → #740 (v8.2) → #736 (v8.1).

Adds an opt-in buildx registry-side layer cache. Off by default — turn on with `cache: true` on `push-to-registries`.

When enabled, buildx writes layers to `<image>:buildcache` on the first pushed registry (`mode=max`, all intermediate layers cached) and reads from it on subsequent builds. Same registry as the image, separate tag — no extra repo to manage. The cache is cross-job-host because it lives in the registry, not in CircleCI's infra.

Different from the existing `go-cache-save`/`go-cache-restore`: those cache Go module downloads in CircleCI's cache; this caches buildx-produced image layers in the registry. Complementary — they help different stages.

Wins are biggest for compile-in-Dockerfile patterns (`RUN go build`, `RUN apk add`); for the prebuilt-binary pattern (`ADD ./bin-${TARGETARCH}` only) the saving is mostly base-image reuse. Default-off so consumers explicitly choose to pay for the extra registry write.


QuentinBisson:
Closing per discussion: no project is currently asking for build-cache help, and for the dominant prebuilt-binary Dockerfile pattern (architect, silence-operator, observability-operator) the cache is net overhead — extra registry writes per build for cache misses on every commit. Re-open when a real project surfaces a slow-build pain point that this would solve.

PR #747, opened 2026-05-07
feat!: breaking cleanups (collapse to single buildx path, drop deprecated)

QuentinBisson:
Target: **v9.0.0**. Top of the stack: #745 (v8.3) → #753 (v8.4 helm sign) → #754 (v8.5 binary sign/upload/hadolint) → #755 (v8.6 verify-after-sign) → **this PR**.

The only PR in the stack with breaking changes — everything else lands as v8 minors first.

### Removed

- `push-to-registries-multiarch` job (deprecated since v7.0). Consumers rename to `push-to-registries`.
- `multiarch:` parameter on `push-to-registries`. Always buildx now; consumers using `multiarch: false` get multi-arch automatically.
- `os` parameter on `go-build` (deprecated, ignored since v8.1).
- Singular `architecture` parameter on `go-build`. Replaced by `architectures` (plural). Matrix-based callers must switch to `architectures: "linux/amd64,linux/arm64"` on a single job.
- `image-build-with-docker` and `image-push-to-registries` commands. The single-arch path is gone.
- Two `[deprecated]` legacy OCI auth/push run steps in `push-helm` plus `password_envar`/`username_envar`/`registry_url` parameters. These were already orphaned — only the deprecated block ever read them; the live OCI push step hardcodes `gsoci`/`gsociprivate` and reads `ACR_GSOCI_*` / `ACR_GSOCIPRIVATE_*` env vars.

### Changed

- `go-build` default `architectures` is now `linux/amd64,linux/arm64` (previously `linux/amd64`). Consumers that relied on the implicit default get multi-arch binaries (and a matching `.platforms` file) automatically; `push-to-registries` picks both up without any extra config. Single-arch consumers set `architectures: "linux/amd64"` explicitly. Dockerfiles that `ADD <binary> /…` (no `${TARGETARCH}` selector) will silently ship the amd64 binary into the arm64 manifest — switch to the `${TARGETARCH}` pattern or pin back to single-arch.

### Renamed

- `image-build-and-push-multiarch` (command) → `image-build-and-push`. Multi-arch is no longer a distinguishing trait. Direct callers of the command need to update; `push-to-registries` job consumers are unaffected.

### Docs

- `push-to-registries.md` and `go-build.md` rewritten for the v9 surface with an explicit Dockerfile contract section: the `${TARGETARCH}` selector pattern is documented as the recommended path; `--platform=$BUILDPLATFORM` compile-in-Dockerfile pattern documented as the alternative.
- `push-to-app-catalog.md` updated to reflect that registry URL / auth are no longer configurable (always gsoci/gsociprivate).
- `migration-v8-to-v9.md` and the v9 defaults table document the new `go-build` default and how to pin back to single-arch.


QuentinBisson:
@piontec are you fine with me mergint this in v9-pre-release?

PR #748, opened 2026-05-08
Sign published Helm charts with cosign keyless OIDC

QuentinBisson:
Target: **v9.1.0**. Stacked on #747 (v9 cleanups) → #745 (v8.3) → #740 → #736.

Extends cosign keyless signing to Helm OCI charts pushed by the `push-to-app-catalog` job. Same mechanism as container-image signing: parse the OCI manifest digest from `helm push` output, mint a CircleCI OIDC token with `aud=sigstore`, sign the chart digest. **Public charts only** — private charts skipped to avoid public Rekor leakage.

`sign: true|false` (default `true`) on `push-to-app-catalog` and the underlying `push-helm` command.

The signing is inlined into the existing helm push step (`if [[ "<<parameters.sign>>" == "true" && "${CHART_ACCESS}" == "public" ]]; then ...`) rather than a separate `when:` block — variables stay in scope, no /tmp file plumbing. (The follow-on PR #752 extracts the token-mint into a shared `cosign-prepare` command.)

Sigstore cert source-repo extension (OID `1.3.6.1.4.1.57264.1.12`) populates `github.com/<org>/<repo>` like the image signatures, so verification policies can pin to the readable identity.

`cosign` is assumed on PATH from the architect image (giantswarm/architect#1314); the inline install was dropped.

### DIFF
--- a/src/commands/push-helm.yaml
+++ b/src/commands/push-helm.yaml
@@ -234,16 +234,7 @@ steps:
                 chart_ref="${OCI_REGISTRY_URL}/charts/giantswarm/<<parameters.chart>>@${chart_digest}"
                 echo "Signing chart: $chart_ref"
 
-                # renovate: datasource=github-releases depName=sigstore/cosign
-                COSIGN_VERSION=v3.0.3
-                host_arch=$(uname -m)
-                case "$host_arch" in
-                  x86_64)  cosign_arch=amd64 ;;
-                  aarch64) cosign_arch=arm64 ;;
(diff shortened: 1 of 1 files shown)


piontec (commented):
Great one, with minor ideas to improve:
- let's skip (here and in image signing) the dowload of `cosign` and embed it into our runner image. This really means hundreds of downloads a day of a binary you know you need anyway.
- let's include this already in 8.x

QuentinBisson:
Image update is here https://github.com/giantswarm/architect/pull/1314

PR #749, opened 2026-05-11
Verify each cosign signature immediately after producing it

QuentinBisson:
Target: **v9.2.0**. Stacked on #752 (v9.4 cosign-prepare/binary signing/release upload/hadolint) → #748 (v9.1 helm cosign) → #747 (v9) → #745 → #740 → #736.

Adds `cosign verify` right after every `cosign sign` in:

- `image-build-and-push` (per registry, per image)
- `push-helm` (per chart)

Both call with regex-based identity assertions matching CircleCI's UUID-based SAN URI:

```
cosign verify \
  --certificate-oidc-issuer-regexp '^https://oidc\.circleci\.com' \
  --certificate-identity-regexp '^https://circleci\.com/api/v2/projects/[a-f0-9-]+/pipeline-definitions/[a-f0-9-]+$' \
  "$ref"
```

The `--certificate-source-repository-uri` flag (which would map to Sigstore OID `1.3.6.1.4.1.57264.1.12` — the friendly `github.com/<org>/<repo>` extension) doesn't exist in cosign v3.0.3. The cert *does* carry the value, but cosign's CLI surface doesn't expose it without a `--policy` rego file. The SAN regex still catches: invalid signature chain, wrong OIDC issuer, wrong SAN shape. Stronger repo-pinning awaits cosign exposing the newer OIDs.

**Why this matters**: catches silent regressions where a `cosign sign` returns success but the resulting signature wouldn't verify (wrong audience, wrong issuer, identity mismatch). Without verify-after-sign, we ship "signed" artifacts consumers can't verify and only find out from a consumer report.


piontec (approved):
Let's move it to 8.v release

QuentinBisson:
I'll build this on the cosign branch instead then

PR #750, opened 2026-05-11
v9.4: cosign-prepare extraction + binary signing + release upload + hadolint

QuentinBisson:
## Summary

Four supply-chain additions stacked on top of #v9.2 (verify-after-sign):

1. **`cosign-prepare` command.** Single shared step that installs cosign if missing (no-op once the architect image bakes it in via giantswarm/architect#1314) and mints the sigstore-audience CircleCI OIDC token. Replaces the duplicated install+mint blocks in `image-build-and-push` and `push-helm`.
2. **Cosign keyless signing of Go binaries** in `go-build`. Each `<binary>-<GOOS>-<GOARCH>` gets a sibling `.bundle` Sigstore bundle, immediately verified with `cosign verify-blob`. Public repos only — private repos skip at runtime to avoid leaking artifacts into public Rekor. Controlled by `sign: true|false` (default `true`).
3. **`upload-release-assets` job.** Attaches workspace-persisted binaries (and their `.bundle` siblings) to the GitHub Release for the current tag via `gh release upload --clobber`, retrying up to `attempts` × 5s to absorb the race with the create-release GitHub Action. Tag-filtered. Helm charts and container images stay in OCI.
4. **`hadolint` step** in `push-to-registries`, gated by `hadolint: warn|fail|skip` (default `warn`) and `hadolint-config`. Uses the `/usr/bin/hadolint` baked into the architect image.

Stacked on #N/A (feat/v9.2-verify-after-sign). Set this PR's base to `main` after v9.2 merges.

## Notes

- `cosign-prepare` is intentionally tiny; the install half becomes a no-op the moment giantswarm/architect#1314 lands and ships an architect image with cosign on PATH. The token mint stays, because the auto-injected `CIRCLE_OIDC_TOKEN_V2` has the wrong audience for Fulcio's CircleCI federation.
- Binary bundles ride through the existing `./<binary>-*-*` workspace glob (the trailing `-*` matches `<GOARCH>.bundle`), so no jobs/go-build persist-path changes were needed.
- `hadolint` defaults to `warn` so it lights up findings without breaking existing pipelines. Projects opt into hard-fail per repo.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- New `cosign-prepare` command. Mints a CircleCI OIDC token with `aud=sigstore` via `circleci run oidc get --root-issuer`, exporting it as `SIGSTORE_ID_TOKEN` through `BASH_ENV`. The cosign signing blocks in `image-build-and-push` and `push-helm` now call this single command instead of duplicating the token-mint logic.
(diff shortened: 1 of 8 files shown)


piontec (changes requested):
Let's already drop the `cosign` installation path and use the image with bin.

PR #158, opened 2020-05-24
Retry creating kind cluster if fails

fiunchinho:
## Checklist

- [X] Make yourself familiar with following readme sections:
    - [X] [Coding Standards](https://github.com/giantswarm/architect-orb#coding-guidelines).
    - [X] [Development](https://github.com/giantswarm/architect-orb#development).
    - [X] [Design and Goals](https://github.com/giantswarm/architect-orb#design-and-goals).
- [X] :warning: Update changelog in [CHANGELOG.md](https://github.com/giantswarm/architect-orb/tree/master/CHANGELOG.md).
- [ ] :warning: After the release update architect orb version in [.circleci/config.yml](https://github.com/giantswarm/architect-orb/tree/master/.circleci/config.yml).

PR #122, opened 2020-04-09
Update Go version used in integration test to 1.14.1

rossf7:
## Checklist

- [x] Make yourself familiar with following readme sections:
    - [x] [Coding Standards](https://github.com/giantswarm/architect-orb#coding-guidelines).
    - [x] [Development](https://github.com/giantswarm/architect-orb#development).
    - [x] [Design and Goals](https://github.com/giantswarm/architect-orb#design-and-goals).
- [x] :warning: Update changelog in [CHANGELOG.md](https://github.com/giantswarm/architect-orb/tree/master/CHANGELOG.md).
- [ ] :warning: After the release update architect orb version in [.circleci/config.yml](https://github.com/giantswarm/architect-orb/tree/master/.circleci/config.yml).

PR #48, opened 2019-10-31
add "tag-latest-branch" parameter to the push-to-docker job

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/7480

I figured that when architect image is build it's pushed to quay using version from `architect project version` command but the `latest` tag isn't updated. The job is updated to also push `latest` image on (by default) `master` branch.

PR #771, opened 2026-05-19
fix(cosign-prepare): self-install cosign on executors that don't ship it

### LINKED ISSUE #769: cosign step in push-to-app-catalog breaks executor: app-build-suite
## Problem

Since v8.2.0, `architect/push-to-app-catalog` defaults `sign: true` and runs `cosign sign` / `cosign verify` from whatever executor was selected. The `cosign-prepare` command unconditionally invokes `cosign version` as its very first step.

When a consumer selects `executor: app-build-suite` (so the `app_build_suite` Python CLI is available to package the chart with metadata, lint, kubeconform, etc. -- the established pattern for repos that ship a Helm app), the job fails on the new `Mint Sigstore OIDC token` step:

```
/bin/bash: line 3: cosign: command not found
Exited with code exit status 127
```

The `gsoci.azurecr.io/giantswarm/app-build-suite:1.8.0-circleci` image does not ship `cosign`.

The `push-to-registries` cosign path is unaffected (it runs on the `architect` executor which has cosign installed).

## Impact

Any repo that uses `executor: app-build-suite` on `push-to-app-catalog*` and uses the default `sign: true` will break on its next chart push (branch or tag, depending on workflow shape). Observed in `giantswarm/klaus` PR `#263` after a one-line orb bump from `8.1.0` to `8.2.1`.

Five repos in the Bumblebee fleet override this executor and are affected:
- `giantswarm/klaus`
- `giantswarm/klaus-gateway`
- `giantswarm/mcp-capi`
- `giantswarm/mcp-kubernetes`
- `giantswarm/mcp-prometheus`

Working around it per-repo with `sign: false` on every `push-to-app-catalog` invocation is what we just did to unblock the orb bump, but it leaves the public charts unsigned -- the exact opposite of what enabling `sign: true` by default in `v8.2.0` was meant to achieve.

## Proposed solution

Either:

1. **Install `cosign` in the `app-build-suite` executor image** (`gsoci.azurecr.io/giantswarm/app-build-suite` -- the upstream image lives outside this repo, so this option requires a coordinated change there). Cleanest; signing then works regardless of which orb job picks the executor.
2. **Make `cosign-prepare` self-install cosign when missing** (download the official static binary into a writable PATH location). Self-contained inside this orb, no dependency on the `app-build-suite` image's release cadence.
3. **Skip `cosign sign` gracefully when `cosign` is not in `PATH`**, with a clear `echo "cosign not found in executor; skipping chart signing"` warning, so consumers don't need to explicitly opt out per-call.

Option 2 (or 3 as a complement) is the most ergonomic for orb consumers because no consumer config edits are required when the bug is fixed -- they just pick up the next orb version and signing turns on.

## Acceptance criteria

- A repo using `executor: app-build-suite` on `architect/push-to-app-catalog` with the default `sign: true` completes the chart push (and the cosign signing step) successfully on both branch and tag pipelines.
- Existing consumers that have set `sign: false` as a workaround can remove that line and pick up signing again with the next orb bump.

## Notes

The five consumer workarounds are currently:
- klaus: PR #263 (sign: false amended into the orb bump)
- klaus-gateway: PR #48 (same)
- mcp-capi: PR #122 (same)
- mcp-kubernetes: PR #424 (separate follow-up after #422 already merged)
- mcp-prometheus: PR #158 (separate follow-up after #156 already merged)

QuentinBisson:
## Summary

Closes #769.

Since `v8.2.0` defaulted `sign: true` on `push-to-app-catalog`, every consumer that uses `executor: app-build-suite` fails on the new `Mint Sigstore OIDC token` step with:

```
/bin/bash: line 3: cosign: command not found
Exited with code exit status 127
```

because the `gsoci.azurecr.io/giantswarm/app-build-suite:1.8.0-circleci` image does not ship `cosign`. The five Bumblebee-fleet repos affected (`klaus`, `klaus-gateway`, `mcp-capi`, `mcp-kubernetes`, `mcp-prometheus`) currently work around this with `sign: false`, which leaves public charts unsigned — the exact opposite of what `v8.2.0`'s default was meant to achieve.

## What changes

`cosign-prepare` now self-installs the official static `cosign` binary into `${HOME}/.local/bin` when `cosign` is not already on `PATH`:

- Detects the executor architecture (`x86_64` → `amd64`, `aarch64`/`arm64` → `arm64`); errors out for anything else.
- Downloads the pinned `cosign-linux-<arch>` from the `sigstore/cosign` GitHub release.
- Verifies its SHA-256 against the upstream `cosign_checksums.txt` (also downloaded from the same release).
- `chmod +x`, then `export PATH=…` via `BASH_ENV` so the existing `Mint Sigstore OIDC token` step and the downstream `cosign sign` / `cosign verify` calls all pick it up.
- Skips entirely when `cosign` is already on `PATH` (the `architect` executor case — no behavior change there).

A new command parameter `cosign_version` (default `3.0.6`) exposes the pin. Renovate tracks it via a new `customManager` in `renovate.json5` (datasource `github-releases`, depName `sigstore/cosign`, same `extractVersionTemplate: '^v(?<version>.+)$'` pattern used for `kyverno_version` and `yq_version`), so the orb stays current without manual edits.

## Effect on consumers

After this ships and Renovate bumps each repo's orb pin, the five workaround PRs can drop the `sign: false` line and pick up keyless chart signing for free:

| Repo | Workaround PR |
|---|---|
| klaus | #263 |
| klaus-gateway | #48 |
| mcp-capi | #122 |
| mcp-kubernetes | #424 |
| mcp-prometheus | #158 |

## Notes

- Pure runtime change inside `cosign-prepare`. All three callers (`push-helm`, `go-build`, `image-build-and-push-multiarch`) inherit the fix without any per-call edit; the new parameter has a default so existing call sites compile as-is.
- Supply chain: the downloaded binary is SHA-256-verified against the official `cosign_checksums.txt` from the same release tag, which is what cosign's own [installation instructions](https://docs.sigstore.dev/system_config/installation/#installation-on-linux-with-the-binary) recommend for environments where Homebrew/the GitHub Action isn't applicable. We do not transitively verify `cosign_checksums.txt` itself with an older cosign (bootstrap-circular).
- The `architect` executor's baked-in `cosign` continues to be used as-is — `command -v cosign` short-circuits before any download.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- `cosign-prepare` now self-installs the official `cosign` static binary into `${HOME}/.local/bin` when `cosign` is not already on `PATH`, with SHA-256 verification against the upstream `cosign_checksums.txt`. The v8.2.0 default of `sign: true` on `push-to-app-catalog` broke every consumer that uses `executor: app-build-suite` (the `gsoci.azurecr.io/giantswarm/app-build-suite:1.8.0-circleci` image ships no `cosign`) with `cosign: command not found` on the new `Mint Sigstore OIDC token` step. Existing consumers that worked around the breakage with `sign: false` can drop that line on the next orb bump and pick up signing again. The `architect` executor path is unchanged — its baked-in `cosign` is detected and reused. Pinned version is exposed via a new `cosign_version` command parameter (default `3.0.6`) and tracked by Renovate (`sigstore/cosign` GitHub releases). Closes #769.
+
 ## [8.2.1] - 2026-05-19
 
 ### Fixed
--- a/renovate.json5
+++ b/renovate.json5
@@ -105,6 +105,18 @@
       datasourceTemplate: 'github-releases',
       extractVersionTemplate: '^v(?<version>.+)$',
     },
+    {
+      customType: 'regex',
+      managerFilePatterns: [
+        '/^src/commands/cosign-prepare\\.yaml$/',
+      ],
+      matchStrings: [
+        'cosign_version:[\\s\\S]*?default:\\s*"(?<currentValue>[^"]+)"',
+      ],
+      depNameTemplate: 'sigstore/cosign',
+      datasourceTemplate: 'github-releases',
+      extractVersionTemplate: '^v(?<version>.+)$',
+    },
   ],
   packageRules: [
     {
--- a/src/commands/cosign-prepare.yaml
+++ b/src/commands/cosign-prepare.yaml
@@ -1,21 +1,90 @@
 # Command: cosign-prepare
 #
-# Mints a CircleCI OIDC token with audience=sigstore via
-# `circleci run oidc get --claims '{"aud":"sigstore"}' --root-issuer` and
-# exports it as SIGSTORE_ID_TOKEN through BASH_ENV so subsequent run steps in
-# the same job can `cosign sign` / `cosign verify` without additional setup.
+# Ensures cosign is on PATH (self-installing the official static binary into a
+# writable location when missing — needed for executors such as `app-build-suite`
+# whose image does not ship cosign), then mints a CircleCI OIDC token with
+# audience=sigstore via `circleci run oidc get --claims '{"aud":"sigstore"}'
+# --root-issuer` and exports it as SIGSTORE_ID_TOKEN through BASH_ENV so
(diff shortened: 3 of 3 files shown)


QuentinBisson:
Closing in favour of installing cosign directly into the `app-build-suite` image (option 1 from #769). Runtime self-install would have worked but is strictly less clean: every `cosign-prepare` invocation pays a download + SHA-256 verify, the orb carries supply-chain logic that belongs in the executor image, and the version pin lives one layer removed from where the binary actually runs. PR coming up against `giantswarm/app-build-suite` instead.

PR #774, opened 2026-05-20
push-helm: derive cosign chart ref from helm push output; fold sign+verify into one command

QuentinBisson:
## Summary

Two related changes after the v8.2.x cosign rollout shook out:

**1. `push-helm` cosign signing now signs the artifact `helm push` actually wrote.**

Consumers that set `explicit_allow_chart_name_mismatch: true` — where the Chart.yaml `name` differs from the orb's `chart` parameter — were failing the new v8.2.x cosign step. The most visible case is `giantswarm/releases`, which uses a static `chart: release-chart` while the actual Chart.yaml inside the tgz says `release-aws` / `release-azure` / etc. The signing block built its reference from `<<parameters.chart>>`, so cosign hit a `404 Not Found` on a non-existent `release-chart@<digest>` manifest (real failure: https://app.circleci.com/pipelines/github/giantswarm/releases/12683/workflows/e0505b86-70e8-4f78-82ca-dae536fb4203/jobs/43507).

`helm push` always tolerated dir-name ≠ Chart.yaml name; the orb shouldn't break that pattern. Parse the `Pushed: oci://…` line from helm push output to recover both the digest and the real OCI repo path, and sign that. The shape mirrors the existing `--metadata-file` capture in `image-build-and-push-multiarch` — same place, same idea.

**2. Cosign `sign` + `verify` blocks (image / chart / binary) collapsed into one command.**

The same five-line sign + verify pair appeared three times across `image-build-and-push-multiarch`, `push-helm`, and `go-build`, each with the CircleCI OIDC identity / issuer regex pair hard-coded inline. New `cosign-sign-verify` command:

- Mints the sigstore-audience OIDC token internally (absorbs the old `cosign-prepare`, which is removed).
- Dispatches on `kind: oci|blob` — `oci` runs `cosign sign` / `cosign verify` on OCI refs; `blob` runs `cosign sign-blob --bundle` / `cosign verify-blob --bundle` on files.
- Reads one entry per line from a `refs_file` parameter (default `/tmp/.cosign_refs`). Empty file = no-op, so callers signal "skip" by truncating.

Callers (push-helm, image-build-and-push-multiarch, go-build) now stage their entries inside their existing shell loops and invoke the command once. Identity regex pair lives in exactly one place.

## Files

- **new** `src/commands/cosign-sign-verify.yaml`
- **deleted** `src/commands/cosign-prepare.yaml` (folded in)
- `src/commands/push-helm.yaml` — derive ref from helm push output; call shared command
- `src/commands/image-build-and-push-multiarch.yaml` — per-registry loop now appends to refs file; call shared command
- `src/commands/go-build.yaml` — binary loop now appends `<file> <bundle>` pairs to refs file; call shared command with `kind: blob`

## Affected downstream

- `giantswarm/releases` — unblocks every cluster-`<provider>` release rebuild (was 100% failing on v8.2.x).
- Any other consumer using `explicit_allow_chart_name_mismatch: true` (rare pattern but legitimate).
- Everyone else: unchanged behaviour, slightly fewer steps per cosign-enabled job (the `cosign-prepare` step is gone, sign+verify now sits in one step).

## Not in this PR

The companion `go-build` issue (renovate-bumped consumers fail at `Generate temporary GitHub token` because they don't pass `context: architect`) is a consumer-side fix — Marco's 8 listed PRs + the rest of the org sweep. Tracking separately.


piontec (approved):
LGTM

PR #778, opened 2026-05-20
feat: add branch-tag for per-branch image and chart tags

marians:
## Summary

Adds an opt-in `branch-tag: bool` parameter (default `false`) to both `push-to-registries` and `push-to-app-catalog` so a single push can produce a co-versioned image + chart pair keyed on the branch.

### `push-to-registries` — image tag stream
- When `true`, pushes `<image>:branch-<slug>-<CIRCLE_BUILD_NUM>` alongside the existing `<semver>-<sha>` and `:latest<suffix>` tags.
- Plumbed through both the single-arch (`image-push-to-registries`) and multi-arch (`image-build-and-push-multiarch`) paths. Same `push_dev` / access-visibility gating as the primary tag. Cosign signing is by index digest (multi-arch) — tag-independent.

### `push-to-app-catalog` — chart tag stream
- When `true`, packages the chart with `helm package --version 0.0.0-branch-<slug>-<CIRCLE_BUILD_NUM>` (valid SemVer pre-release) and pushes to the OCI registry under that tag.
- The legacy git-based app catalog push (`push_to_appcatalog`) is force-skipped: per-branch builds are not meant for the production catalog repo. Only the OCI push runs.
- Requires `executor: architect` — guard fails fast on `app-build-suite` (its packaging step doesn't honor `helm package --version`). Extending app-build-suite support is deferred.

### Shared mechanics
- Slug is `CIRCLE_BRANCH` with `/` replaced by `-`, validated against the DNS-1123 label regex `^[a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?\$`. Job fails early if it doesn't match — catches non-deploy-safe branch names before any image is built. Validation runs only when the parameter is on, so consumers with branch names like `feat/foo` are unaffected unless they opt in.
- Slug computation lives in a single command (`image-compute-branch-slug`) shared by both jobs.

## Why

For per-branch prototype deployments where Flux Image Update Automation needs a numerically-sortable tag stream keyed on the branch (build number is monotonic; SHA isn't sortable; rolling tags can't drive IUA-based reconcile because the manifest text never changes). Co-versioning the image and chart per push lets a per-branch Flux \`OCIRepository\` + \`ImagePolicy\` pair bump them together.

## Test plan

- [x] Dev orb publishes successfully (\`orb-tools/pack\` + \`circleci orb validate\` clean).
- [x] Consumer enables \`branch-tag: true\` on a DNS-1123-compliant branch → image pushed as \`branch-<slug>-<build-num>\` (verified: \`gsociprivate.azurecr.io/giantswarm/agentic-platform-ui:branch-test-branch-tag-dev-orb-70\`).
- [x] Same consumer → chart pushed as \`0.0.0-branch-<slug>-<build-num>\` (verified: \`gsociprivate.azurecr.io/charts/giantswarm/agentic-platform-ui:0.0.0-branch-test-branch-tag-dev-orb-74\`).
- [x] Legacy git appcatalog push skipped when \`branch-tag: true\` (verified: no commit to catalog repo on \`:74\` run).
- [ ] Same consumer on a non-conforming branch (e.g. \`feat/foo\`) fails with a clear error before image build / chart package.
- [ ] Existing consumers (no \`branch-tag\` set) see byte-identical orb behavior — covered by the no-op default and the gated \`when:\` blocks, but a regression run on a current production consumer would be reassuring before merge.
- [ ] Multi-arch image path: index digest is signed (cosign), all tag references resolve to the signed index.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Added
+
+- New `branch-tag` parameter on `push-to-registries` (off by default). When `true`, the job pushes an additional, branch-keyed tag `<image>:branch-<slug>-<CIRCLE_BUILD_NUM>` alongside the standard `<semver>-<sha>` tag, where `<slug>` is `CIRCLE_BRANCH` with `/` replaced by `-`. The slug is validated against the DNS-1123 label regex (`^[a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?$`) and the job fails early if it doesn't match — catches non-path-safe branch names before any image is built. Intended for per-branch deployments where Flux Image Update Automation needs a numerically-sortable tag stream keyed on the branch. Existing callers see no behavior change.
+
 ## [8.3.0] - 2026-05-20
 
 ### Added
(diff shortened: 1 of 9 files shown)


marians:
`giantswarm/architect@dev:6295e54c83dcf84d31f99b326b5c36833d0b45c9` was published

marians:
Tested via https://github.com/giantswarm/agentic-platform-ui/pull/11

marians:
This is obsolete, as Honeybadger will soon deliver a solution for branch-based tags.

PR #793, opened 2026-06-01
feat(push-to-registries): attach per-arch CycloneDX SBOMs to multi-arch images

marians:
Towards https://github.com/giantswarm/giantswarm/issues/36746

## Description

BuildKit's native `--attest type=sbom` only emits **SPDX**. This PR adds an opt-in way to also publish a **CycloneDX** SBOM. A CycloneDX SBOM is generated **per architecture** with [syft](https://github.com/anchore/syft) and attached to each platform manifest as an **unsigned OCI 1.1 referrer** (artifactType `application/vnd.cyclonedx+json`) using [oras](https://oras.land).

Like the inline SPDX SBOM, the CycloneDX SBOM is **unsigned** and attached the same way for **both public and private images** — no cosign, no Rekor transparency log. (An earlier iteration used keyless `cosign attest`, but the architect image's cosign rejects `--tlog-upload=false` under a signing config, and a signed attestation isn't needed to match how SPDX is published.)

> Rebased onto **v9.0.0**. v9 collapsed the single-arch/multi-arch split into one `docker buildx` path and renamed the `image-build-and-push-multiarch` command to `image-build-and-push`; this branch now adds the feature to that unified command.

### Changes

**New parameter (default `false`, off):**
- `push-to-registries` job and `image-build-and-push` command: `sbom-cyclonedx`. When `true`, generates and attaches per-arch CycloneDX SBOMs after the image push.

**Behavior:**
- SBOMs are generated **per architecture** by scanning each pushed platform manifest by digest, then attached as referrers to that same digest.
- Runs for **both public and private images** (unsigned, no transparency log).
- Attachment happens **after the push** — an OCI referrer must point at an already-published manifest (same reason cosign signing runs post-push).
- Respects the `split-china-push` and `push_dev` filters; dev-tag detection uses `gitsemver validate --type dev`.
- `syft` and `oras` must be present in the architect image; the step fails loudly if either is missing (no on-the-fly install).

**Tooling:**
- Pins the `architect` executor image to the `8.0.0-…` tag that ships `syft` and `oras`.

**Documentation:**
- `docs/job/push-to-registries.md`: CycloneDX SBOM section, including how to list/pull the SBOM via `oras discover` / `oras pull`.
- `CHANGELOG.md`: feature entry under `[Unreleased]`.

### Consuming the SBOM

```sh
# List the CycloneDX SBOM attached to a platform manifest
oras discover --artifact-type application/vnd.cyclonedx+json \
  <registry>/giantswarm/myapp@<platform-digest>

# Pull the SBOM blob
oras pull <registry>/giantswarm/myapp@<sbom-referrer-digest>
```

### Checklist

- [x] Familiar with [Coding Standards](https://github.com/giantswarm/architect-orb#coding-guidelines)
- [x] Familiar with [Development](https://github.com/giantswarm/architect-orb#development)
- [x] Familiar with [Design and Goals](https://github.com/giantswarm/architect-orb#design-and-goals)
- [x] Updated [CHANGELOG.md](https://github.com/giantswarm/architect-orb/tree/master/CHANGELOG.md)

### Test Plan

The feature is opt-in (`sbom-cyclonedx: false` by default). Verified on a real build (public and private images): the per-arch CycloneDX SBOMs are generated by syft and attached to each platform digest, discoverable via `oras discover`.

https://claude.ai/code/session_012dzEgfdM39XoMHfdkzYEiS


src/commands/image-build-and-push-multiarch.yaml
```diff
@@ -343,3 +353,86 @@ steps:
...
+                exit 1
+              fi
+
+              # syft turns the pushed per-arch image into a CycloneDX SBOM. Pinned and
+              # bumped by Renovate; installed only if the architect image doesn't ship it.
+              # renovate: datasource=github-releases depName=anchore/syft
+              SYFT_VERSION=v1.44.0
+              if ! command -v syft >/dev/null 2>&1; then
+                echo "syft not found on PATH; installing ${SYFT_VERSION}."
+                curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh \
```
piontec: Please don't do that - don't install dependcies here. We run this hundreds of times per day, we want to have necessay binaries in the image and not to install them on the fly. I know this is a backup, but it's a backup that will hide the real problem if it exists. Just make the dep in the image and fail here if it's not available.

src/commands/image-build-and-push-multiarch.yaml
```diff
@@ -343,3 +353,86 @@ steps:
...
+                  | sh -s -- -b /usr/local/bin "${SYFT_VERSION}"
+              fi
+              syft version
+
+              TAG_VALUE="${DOCKER_IMAGE_TAG}"
+              while IFS=' ' read -r access reg _ _ push_dev; do
+                [[ -z "$reg" || "$reg" =~ ^# ]] && continue
+                [[ "<<parameters.split-china-push>>" == "true" && "${reg}" == "giantswarm-registry.cn-shanghai.cr.aliyuncs.com" ]] && continue
+                if [[ "$push_dev" == "false" && "$TAG_VALUE" =~ [a-f0-9]{40} ]]; then
```
piontec: This is no longer the valid way to test if tag is of certain kind. Use `gitsemver validate --type dev`

piontec (changes requested):

src/commands/image-build-and-push-multiarch.yaml
marians: That's https://github.com/giantswarm/gitsemver

src/commands/image-build-and-push-multiarch.yaml
marians: Done in bf0ccb0 — removed the on-the-fly install (and the syft Renovate entry); the step now fails with a clear error if `syft` isn't on `PATH`. Heads-up: this means `syft` needs to be added to the architect image before `sbom-cyclonedx: true` can be used, otherwise the job will fail at this check.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/commands/image-build-and-push-multiarch.yaml
marians: Done in bf0ccb0 — the CycloneDX block now uses `gitsemver validate --type dev "$TAG_VALUE"` instead of the SHA regex. Scoped to just this new block as agreed; the pre-existing checks in the main push loop, the cosign-sign staging, and the single-arch `image-push-to-registries.yaml` still use the old regex and can be modernized separately if you'd like.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/commands/image-build-and-push-multiarch.yaml
marians: Adding syft to architect here: https://github.com/giantswarm/architect/pull/1341

src/executors/architect.yaml
```diff
@@ -1,4 +1,4 @@
 docker:
   - entrypoint: /bin/bash
     # registry: gsoci.azurecr.io/giantswarm/architect
-    image: gsoci.azurecr.io/giantswarm/architect:8.0.0
+    image: gsoci.azurecr.io/giantswarm/architect:8.0.0-e4a3a8e0a0a5d15c375a762bfbdf3a3337224f2d
```
marians: TODO: set release version once https://github.com/giantswarm/architect/pull/1341 is merged and released

src/commands/image-build-and-push.yaml
```diff
@@ -343,3 +353,83 @@ steps:
...
+                  plat_ref="${base}@${platform_digest}"
+                  safe=$(echo "${reg}_${platform}" | tr '/:.' '___')
+                  sbom_file="/tmp/sbom.${safe}.cdx.json"
+                  echo "Generating CycloneDX SBOM for $plat_ref ($platform)"
+                  syft "$plat_ref" -o cyclonedx-json="$sbom_file"
+                  echo "Attaching CycloneDX SBOM to $plat_ref"
+                  # --disable-path-validation: the SBOM lives at an absolute /tmp path,
+                  # which oras rejects by default (the ':' file:mediaType separator is
+                  # ambiguous with absolute paths). Safe here — we control the path.
+                  oras attach --artifact-type application/vnd.cyclonedx+json \
```
piontec, line 426: FYI: Nice, `attach` is the way to go and the new standard, but AFAIR it was only included in OCI 1.1 and some tools still don't support it, ie. `cosign` uses a tagging trick to make their signatures related to objects, but they don't support 'attach' references.

piontec (approved):
TBH, this looks very hacky for "just get a SBOM in a different format", but if it works and there's no tool to "just" confirm them, then it's LGTM

piontec (commented):
Wait, I see a problem: that cyclonedx is not signed with `cosign`, so it's kind of useless, as we have an SBOM, but we don't know if we can trust its content

piontec (changes requested):

piontec (approved):

marians:
Both types of SBOMs we create are not signed. I created https://github.com/giantswarm/giantswarm/issues/36765 for this.

PR #799, opened 2026-06-02
feat(push-to-registries): sign SBOM attestations with cosign keyless OIDC

piontec:
## What

Sign per-platform **SBOM attestations** with cosign keyless OIDC so SBOMs become trustable, verifiable proof of image contents — closing the integrity-vs-authenticity gap described in giantswarm/giantswarm#36765. Reuses the same tooling/patterns already used to sign images, charts, and binaries (sigstore-audience OIDC token, verify-after-sign, public-only gating).

Run validated here: https://app.circleci.com/pipelines/github/giantswarm/app-operator/6666/workflows/7687081b-6159-479d-8f9b-494fb6f5775f/jobs/35945

## Behavior (per platform manifest)

| | `sbom` (SPDX) | `sbom-cyclonedx` |
|---|---|---|
| **Public + `sign`** | signed `cosign attest --type spdxjson` | signed `cosign attest --type cyclonedx` |
| **Private or `sign:false`** | inline buildx SPDX only (unsigned) | unsigned `oras attach` referrer (unchanged from #793) |

- **SPDX** is signed from the **exact predicate buildx already shipped** — extracted via `.predicate` from the inline in-toto statement, *not* regenerated — so the signed copy is byte-identical to the SBOM on the image.
- **CycloneDX** on public images is now signed instead of `oras attach`ed; private images keep the unsigned referrer (signing private artifacts would leak digests/timestamps into the public Rekor log).
- Each signature is verified immediately after signing (`cosign verify-attestation`) with the same CircleCI OIDC issuer/identity assertions consumers use.
- **SLSA provenance signing is intentionally out of scope** (tracked as a follow-up).

## Changes

- `cosign-sign-verify`: new `attest` kind (`cosign attest --type … --predicate` + `cosign verify-attestation`), file format `<ref> <type> <predicate-file>`; no-ops on empty file like the other kinds.
- `image-build-and-push`: generate + stage per-arch SBOMs, sign on public / attach unsigned on private; followed by `cosign-sign-verify: {kind: attest}`. `when` runs on `sbom-cyclonedx OR (sbom AND sign)`.
- `push-to-registries` job: matching parameter description updates.
- Docs: consumer `cosign verify-attestation` procedure (per-platform) + storage tables in `docs/cosign-signing.md` and `docs/job/push-to-registries.md`.
- `CHANGELOG.md`.

## Verification

- CI: `orb-tools/lint` + `pack` validate structure (YAML parses locally).
- On a real public build: `cosign verify-attestation --type spdxjson|cyclonedx …@<platform-digest>` succeeds; a tampered/substituted SBOM fails. Commands documented in `docs/cosign-signing.md`.

Refs giantswarm/giantswarm#36765

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Added
 
-- `push-to-registries`: new `sbom-cyclonedx` parameter (default `false`). When enabled, generates a CycloneDX SBOM **per architecture** with syft and attaches it to each platform manifest as an unsigned OCI 1.1 referrer (artifactType `application/vnd.cyclonedx+json`) using oras. BuildKit's `--attest type=sbom` only emits SPDX, so CycloneDX is produced out-of-band. Unsigned and attached the same way for both public and private images, mirroring the inline SPDX SBOM — no cosign, no Rekor transparency log. Off by default, so existing consumers are unaffected. Requires `syft` and `oras` in the architect image.
+- `push-to-registries`: **sign SBOM attestations** with cosign keyless OIDC on public images, so SBOMs become trustable, verifiable proof of image contents (not just unsigned metadata). For each platform manifest:
+  - **SPDX** (`sbom: true`, default) — the *exact* SPDX predicate buildx produced is extracted from its inline in-toto attestation and re-signed as a cosign DSSE attestation (`cosign attest --type spdxjson`). The signed predicate is byte-identical to the inline one (extracted via `.predicate`, not regenerated), so the signature covers the SBOM that actually ships on the image.
+  - **CycloneDX** (`sbom-cyclonedx: true`) — the syft-generated SBOM is signed as a cosign DSSE attestation (`cosign attest --type cyclonedx`).
+  - Each signature is verified immediately after signing (`cosign verify-attestation`) with the same CircleCI OIDC issuer/identity assertions consumers use.
+  - Signing is **public-only**, consistent with image/chart/binary signing — private images would leak digests/timestamps into the public Rekor transparency log.
+  - SLSA provenance signing is intentionally out of scope for now (tracked separately).
+- `push-to-registries`: new `sbom-cyclonedx` parameter (default `false`). When enabled, generates a CycloneDX SBOM **per architecture** with syft (BuildKit's `--attest type=sbom` only emits SPDX). On public images with `sign: true` it is signed as described above; on private images (or `sign: false`) it falls back to an **unsigned** OCI 1.1 referrer (artifactType `application/vnd.cyclonedx+json`) attached with oras. Off by default, so existing consumers are unaffected. Requires `syft` and `oras` in the architect image.
+
+### Changed
+
+- `cosign-sign-verify`: new `attest` kind alongside `oci`/`blob`. Reads a `<ref> <type> <predicate-file>` file and runs `cosign attest` + `cosign verify-attestation` for signed SBOM/in-toto attestations, keeping the CircleCI OIDC issuer/identity regex pair in one place.
 
 ## [9.0.1] - 2026-06-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,17 +10,25 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Added
 
 - `push-to-registries`: **sign SBOM attestations** with cosign keyless OIDC on public images, so SBOMs become trustable, verifiable proof of image contents (not just unsigned metadata). For each platform manifest:
-  - **SPDX** (`sbom: true`, default) — the *exact* SPDX predicate buildx produced is extracted from its inline in-toto attestation and re-signed as a cosign DSSE attestation (`cosign attest --type spdxjson`). The signed predicate is byte-identical to the inline one (extracted via `.predicate`, not regenerated), so the signature covers the SBOM that actually ships on the image.
+  - **SPDX** (`sbom: true`, default) — the *exact* SPDX predicate buildx produced is extracted from its inline in-toto attestation and re-signed as a cosign DSSE attestation (`cosign attest --type spdxjson`). The signed predicate is the same SBOM content as the inline one (extracted via `.predicate`, not regenerated; semantically identical, though re-serialized by jq/cosign so not byte-for-byte equal), so the signature covers the SBOM that actually ships on the image.
   - **CycloneDX** (`sbom-cyclonedx: true`) — the syft-generated SBOM is signed as a cosign DSSE attestation (`cosign attest --type cyclonedx`).
   - Each signature is verified immediately after signing (`cosign verify-attestation`) with the same CircleCI OIDC issuer/identity assertions consumers use.
   - Signing is **public-only**, consistent with image/chart/binary signing — private images would leak digests/timestamps into the public Rekor transparency log.
   - SLSA provenance signing is intentionally out of scope for now (tracked separately).
+
+### Changed
+
+- `cosign-sign-verify`: new `attest` kind alongside `oci`/`blob`. Reads a `<ref> <type> <predicate-file>` file and runs `cosign attest` + `cosign verify-attestation` for signed SBOM/in-toto attestations, keeping the CircleCI OIDC issuer/identity regex pair in one place.
+- `push-to-registries`: **behavior change for public CycloneDX SBOMs.** On public images with `sign: true`, the CycloneDX SBOM (`sbom-cyclonedx: true`) is now a *signed* cosign attestation (a sigstore-bundle referrer, `application/vnd.dev.sigstore.bundle.v0.3+json`) rather than the unsigned `application/vnd.cyclonedx+json` OCI referrer that v9.0.2 attached. Consumers discovering it via `oras discover --artifact-type application/vnd.cyclonedx+json` will no longer find it on public images — verify it with `cosign verify-attestation --type cyclonedx` instead (see [docs/cosign-signing.md](docs/cosign-signing.md)). Private images (and `sign: false`) are unchanged: still an unsigned `application/vnd.cyclonedx+json` referrer.
 
 ## [9.0.2] - 2026-06-02
 
 ### Changed
 
 - Bumped `gitsemver` to v2.0.0
+
+### Added
+
 - `push-to-registries`: new `sbom-cyclonedx` parameter (default `false`). When enabled, generates a CycloneDX
   SBOM **per architecture** with syft and attaches it to each platform manifest as an unsigned OCI 1.1
   referrer (artifactType `application/vnd.cyclonedx+json`) using oras. BuildKit's `--attest type=sbom` only
--- a/docs/cosign-signing.md
+++ b/docs/cosign-signing.md
@@ -25,9 +25,11 @@ In addition, `push-to-registries` produces:
   **inline** too. On **public** images with `sign: true`, the *exact* SPDX
   predicate buildx produced is additionally extracted and re-signed as a
   cosign keyless attestation (`--type spdxjson`) so it can be verified
-  independently of the registry. The signed predicate is byte-identical to
-  the inline one — we extract `.predicate` from the buildx in-toto statement
-  rather than regenerating the SBOM.
+  independently of the registry. The signed predicate is the same SBOM
+  content as the inline one — we extract `.predicate` from the buildx in-toto
+  statement rather than regenerating the SBOM. It is semantically identical
+  but re-serialized (jq and cosign re-marshal the JSON), so the two copies are
+  not byte-for-byte equal; compare predicate content, not raw bytes.
 
 BuildKit does not yet support OCI 1.1 referrer storage for its own
 attestations, which is why the inline copies remain; `docker buildx
@@ -136,6 +138,12 @@ cosign verify-attestation \
 signing identity matches; a substituted or tampered SBOM fails. To read the
 verified SBOM payload, pipe the output through
 `jq -r '.payload | @base64d | fromjson | .predicate'`.
+
+`verify-attestation` prints one DSSE envelope per matching attestation, and
+attestations accumulate when the same content-addressed digest is attested
+more than once (e.g. a rebuild of the same image). So that `jq` can emit
+several concatenated predicates — append `| jq -s 'last'` (or otherwise
+select) to pick a single one.
 
 ## Identity model and the friendly-source-repo OID
 
--- a/docs/cosign-signing.md
+++ b/docs/cosign-signing.md
@@ -12,17 +12,27 @@ design.
 | Container image (manifest index) | `push-to-registries` | OCI 1.1 referrer tag on the image | `sign: true` |
 | Helm chart (OCI) | `push-to-app-catalog`, `push-helm` | OCI 1.1 referrer tag on the chart | `sign: true` |
 | Go binary | `go-build` | Sibling `<binary>-<GOOS>-<GOARCH>.bundle` Sigstore bundle | `sign: true` |
+| SPDX SBOM attestation (per platform) | `push-to-registries` | Signed cosign DSSE attestation (OCI referrer) on each platform manifest | `sign && sbom` |
+| CycloneDX SBOM attestation (per platform) | `push-to-registries` | Signed cosign DSSE attestation (OCI referrer) on each platform manifest | `sign && sbom-cyclonedx` |
 
 In addition, `push-to-registries` produces:
 
 - **SLSA provenance** attestation (`provenance: min` by default; `max` is
-  opt-in because it embeds `--build-arg` values verbatim).
-- **SPDX SBOM** attestation (`sbom: true` by default).
+  opt-in because it embeds `--build-arg` values verbatim). Stored **inline**
+  in the image index (as the `unknown/unknown` manifest entries you see in
+  `docker manifest inspect`); currently **unsigned**.
+- **SPDX SBOM** attestation (`sbom: true` by default). buildx stores it
+  **inline** too. On **public** images with `sign: true`, the *exact* SPDX
+  predicate buildx produced is additionally extracted and re-signed as a
+  cosign keyless attestation (`--type spdxjson`) so it can be verified
+  independently of the registry. The signed predicate is byte-identical to
+  the inline one — we extract `.predicate` from the buildx in-toto statement
+  rather than regenerating the SBOM.
 
-Both are stored **inline** in the image index (as the `unknown/unknown`
-manifest entries you see in `docker manifest inspect`) because BuildKit
-does not yet support OCI 1.1 referrer storage for attestations.
-`docker buildx imagetools inspect` displays them with proper labels.
+BuildKit does not yet support OCI 1.1 referrer storage for its own
+attestations, which is why the inline copies remain; `docker buildx
+imagetools inspect` displays them with proper labels. The cosign-signed
+copies use cosign's own referrer push path.
 
 ## Public-only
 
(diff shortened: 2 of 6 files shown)


QuentinBisson:
Solid PR. The extract-don't-regenerate approach for SPDX (pulling `.predicate` from the buildx in-toto statement rather than regenerating) is the right call for byte-identity, and the public-only signing guard is correctly applied in both the condition gate and the `SIGN_SBOMS` logic.

One small thing worth fixing before merge:

**`image-build-and-push.yaml`, SPDX extraction:**

```bash
oras blob fetch --output - "${base}@${spdx_blob}" | jq '.predicate' > "$spdx_file"
```

`jq '.predicate'` should be `jq -e '.predicate'`. Without `-e`, a missing or null `.predicate` silently writes the literal string `null` to the staging file, and `cosign attest` signs it without complaint. The failure only surfaces later during `verify-attestation` with a confusing error. With `-e`, jq exits non-zero on null/false output and the pipeline fails at the source.

src/commands/image-build-and-push.yaml
```diff
@@ -394,12 +415,32 @@ steps:
...
+              if [[ "$SIGN_SBOMS" != "true" && "<<parameters.sbom-cyclonedx>>" != "true" ]]; then
+                echo "No SBOM work: signing skipped (private or sign=false) and CycloneDX not requested."
+                exit 0
+              fi
+
               TAG_VALUE="${DOCKER_IMAGE_TAG}"
               while IFS=' ' read -r access reg _ _ push_dev; do
                 [[ -z "$reg" || "$reg" =~ ^# ]] && continue
                 [[ "<<parameters.split-china-push>>" == "true" && "${reg}" == "giantswarm-registry.cn-shanghai.cr.aliyuncs.com" ]] && continue
                 if [[ "$push_dev" == "false" ]] && gitsemver validate --type dev "$TAG_VALUE" >/dev/null 2>&1; then
```
marians: **[bug]** Dev-tag detection diverges from the push loop this comment claims to mirror.

The push loop (line 198) and cosign-refs loop (line 350) skip `push_dev=false` registries via `[[ "$TAG_VALUE" =~ -dev\. ]]`, but this loop uses `gitsemver validate --type dev`. gitsemver's dev validator is anchored (`^v?X.Y.Z-dev.<branch>.<date>.<time>$`), so a `tag-suffix` (public job parameter, appended by `image-prepare-tag`) or a branch name containing a dot makes the regex match while gitsemver rejects: the push loop skips the registry, this loop proceeds, and `docker buildx imagetools inspect` fails under `set -euo pipefail` against a never-pushed manifest → whole job fails. In the opposite disagreement direction, a pushed image silently gets no SBOM attestation.

Since this step now runs by default (`sbom` + `sign` default `true`), exposure is much wider than the old CycloneDX-only path. Suggest using the identical `-dev\.` regex here (or one shared check for all three loops) so the "filters mirror the push loop" comment above holds.

CHANGELOG.md
```diff
@@ -7,14 +7,20 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+  - **CycloneDX** (`sbom-cyclonedx: true`) — the syft-generated SBOM is signed as a cosign DSSE attestation (`cosign attest --type cyclonedx`).
+  - Each signature is verified immediately after signing (`cosign verify-attestation`) with the same CircleCI OIDC issuer/identity assertions consumers use.
+  - Signing is **public-only**, consistent with image/chart/binary signing — private images would leak digests/timestamps into the public Rekor transparency log.
+  - SLSA provenance signing is intentionally out of scope for now (tracked separately).
+
 ## [9.0.2] - 2026-06-02
 
 ### Changed
 
 - Bumped `gitsemver` to v2.0.0
```
marians, line 28: **[bug]** This PR removes the `### Added` heading from the already-released 9.0.2 section, so the `sbom-cyclonedx` feature bullet below now reads as a "Changed" entry — rewriting released history.

```suggestion
- Bumped `gitsemver` to v2.0.0

### Added

```

src/commands/image-build-and-push.yaml
```diff
@@ -416,20 +457,69 @@ steps:
...
+                    echo "Staging signed SPDX attestation for $plat_ref ($platform)"
+                    echo "${plat_ref} spdxjson ${spdx_file}" >> /tmp/.cosign_attest
+                  fi
+
+                  # --- CycloneDX: syft generates it; sign on public, oras-attach otherwise ---
+                  if [[ "<<parameters.sbom-cyclonedx>>" == "true" ]]; then
+                    cdx_file="/tmp/sbom.${safe}.cdx.json"
+                    echo "Generating CycloneDX SBOM for $plat_ref ($platform)"
+                    syft "$plat_ref" -o cyclonedx-json="$cdx_file"
+                    if [[ "$SIGN_SBOMS" == "true" ]]; then
```
marians, line 513: **[behavior change]** On public images with `sign: true`, the CycloneDX SBOM is no longer attached as an `application/vnd.cyclonedx+json` oras referrer — the signed copy lands as a sigstore bundle referrer (`application/vnd.dev.sigstore.bundle.v0.3+json`; verified live against an image built from this branch). Anyone using the v9.0.2-documented `oras discover --artifact-type application/vnd.cyclonedx+json` on a public image now silently finds nothing.

It's intentional and the docs were updated, but it breaks the just-released contract without a "Changed" CHANGELOG note. Consider either attaching the unsigned referrer *in addition* on public images, or calling the break out explicitly in the CHANGELOG.

src/commands/image-build-and-push.yaml
```diff
@@ -416,20 +457,69 @@ steps:
...
+                      | select(.annotations["in-toto.io/predicate-type"] == "https://spdx.dev/Document")
+                      | .digest' /tmp/.attest_manifest.json | head -n1)
+                    if [[ -z "$spdx_blob" ]]; then
+                      echo "ERROR: no SPDX layer in buildx attestation manifest for $plat_ref." >&2
+                      exit 1
+                    fi
+                    spdx_file="/tmp/sbom.${safe}.spdx.json"
+                    # The blob is an in-toto Statement; cosign attest re-wraps the
+                    # predicate in its own signed Statement, so we hand it just .predicate
+                    # — the SPDX document — keeping the signed copy byte-identical to the
```
marians: **[docs]** "byte-identical" is not literally true — empirically checked against `app-operator` built from this branch: inline statement 464,919 bytes vs signed statement 473,379 bytes. `jq '.predicate'` pretty-prints/renormalizes, and `cosign attest` re-marshals the statement. The predicates *are* canonically identical (sorted-key JSON compare passes), which is the property that matters.

Suggest rewording here, at `docs/cosign-signing.md:28`, and in `CHANGELOG.md:13` to something like "the same SBOM content (semantically identical; re-serialized)" so consumers don't byte-compare the two and report a mismatch.

src/commands/image-build-and-push.yaml
```diff
@@ -416,20 +457,69 @@ steps:
...
+                    # one buildx shipped inline.
+                    oras blob fetch --output - "${base}@${spdx_blob}" | jq '.predicate' > "$spdx_file"
```
marians: **[hardening]** If the fetched blob is JSON without a `.predicate` key, `jq '.predicate'` emits the literal `null` with exit 0, and the pipeline stages and signs a bogus null predicate — `verify-attestation` still passes (it checks signature/identity/type, not predicate content). One flag closes it under `set -euo pipefail`:

```suggestion
                    oras blob fetch --output - "${base}@${spdx_blob}" | jq -e '.predicate' > "$spdx_file"
```

(`jq -e` exits non-zero when the output is `null`/`false`.)

src/commands/cosign-sign-verify.yaml
```diff
@@ -84,4 +91,17 @@ steps:
                 "$file" > /dev/null
             done < "$refs_file"
             ;;
+          attest)
+            while IFS=' ' read -r ref type predicate; do
+              [[ -z "$ref" || "$ref" =~ ^# ]] && continue
+              echo "cosign attest --type $type $ref (predicate: $predicate)"
+              cosign attest --yes --type "$type" --predicate "$predicate" "$ref"
```
marians, line 105: **[fragility note]** Large SBOMs only work here because the architect image's cosign writes new-format sigstore bundles (v0.3), whose Rekor `dsse` entries upload hashes rather than payloads — verified live with a 640 KB bundle that verified including tlog claims. The legacy `intoto` upload path enforces public Rekor's 100 KB attestation limit (sigstore/rekor README), and real SBOMs routinely exceed it (the live app-operator SPDX is ~465 KB; there are public reports of `cosign attest` → `oras attach` migrations for exactly this reason).

If the pinned cosign ever flips bundle formats (or `--new-bundle-format=false` sneaks in), every large-SBOM build starts failing in this loop. Worth a comment documenting the dependency.

src/commands/image-build-and-push.yaml
```diff
@@ -380,10 +401,10 @@ steps:
...
 
-              # syft generates the SBOM and oras attaches it. Both must be provided by the
-              # architect image; fail loudly rather than installing on the fly (this
-              # command runs many times a day, and a fallback install would hide a
-              # missing-dependency regression in the image).
+              # syft generates CycloneDX, oras reads/attaches SBOM blobs. Both must be
+              # provided by the architect image; fail loudly rather than installing on the
+              # fly (this command runs many times a day, and a fallback install would hide
+              # a missing-dependency regression in the image).
               for tool in syft oras; do
```
marians: **[ordering]** This tool check (plus `syft version` / `oras version`) runs *before* the early no-op exit at line 433, and the step now runs by default on every build — including private repos with zero SBOM work. Also, the SPDX-only signing path (`sbom-cyclonedx: false`, the default) never invokes syft yet hard-requires it.

Suggest moving the no-op exit above the tool check, and requiring syft only when `<<parameters.sbom-cyclonedx>>` is true.

src/commands/image-build-and-push.yaml
```diff
@@ -416,20 +457,69 @@ steps:
...
+                  # --- SPDX: sign the exact predicate buildx already attached ---
+                  if [[ "$SIGN_SBOMS" == "true" && "<<parameters.sbom>>" == "true" ]]; then
+                    # buildx links its SPDX attestation manifest to this platform manifest
+                    # via the vnd.docker.reference.* annotations on the index entry.
+                    attest_digest=$(jq -r --arg d "$platform_digest" '
+                      .manifests[]
+                      | select(.annotations["vnd.docker.reference.type"] == "attestation-manifest"
+                               and .annotations["vnd.docker.reference.digest"] == $d)
+                      | .digest' /tmp/.index.json | head -n1)
+                    if [[ -z "$attest_digest" ]]; then
```
marians: **[design]** This `exit 1` (and the one at line 479) fires *after* the image was pushed and its index already cosign-signed, so a missing attestation manifest produces a red build with partially-signed, already-published artifacts — on the default path that v9.0.2 never executed. Live check: ACR does preserve buildx attestation manifests (docker-container driver + `--push`), so this shouldn't fire in current usage — but if buildx storage or a registry's index handling shifts, hard-fail-after-publish is the failure mode. Consider whether warn-and-skip is the better degradation for the missing-manifest case.

src/commands/image-build-and-push.yaml
```diff
@@ -394,12 +415,32 @@ steps:
...
+              : > /tmp/.cosign_attest
+
+              # Nothing to do when we're neither signing nor attaching CycloneDX.
+              if [[ "$SIGN_SBOMS" != "true" && "<<parameters.sbom-cyclonedx>>" != "true" ]]; then
+                echo "No SBOM work: signing skipped (private or sign=false) and CycloneDX not requested."
+                exit 0
+              fi
+
               TAG_VALUE="${DOCKER_IMAGE_TAG}"
               while IFS=' ' read -r access reg _ _ push_dev; do
```
marians: **[cleanup]** Third copy of the registries-data filter scaffold (push loop ~195, cosign-refs ~347, here) — and the copies have already drifted: the dev-check method differs (see comment at line 442) and the cosign loop carries an extra `*public*` gate. That drift is exactly the bug class this duplication breeds. Consider computing the eligible registry rows once (a small shared step writing a file, à la `/tmp/.image_access`) and consuming them in all three loops.

src/commands/image-build-and-push.yaml
```diff
@@ -354,19 +361,33 @@ steps:
...
+              #   CycloneDX — buildx can't emit it, so syft generates it out-of-band.
+              #
+              # On PUBLIC images with sign=true, both are signed as cosign keyless DSSE
+              # attestations (verifiable, trustable provenance — the point of this flow).
+              # On PRIVATE images (or sign=false) signing is skipped (it would leak
+              # digests/timestamps into the public Rekor log), and CycloneDX falls back to
+              # an unsigned oras referrer. The split-china-push / push_dev filters mirror
               # the push loop so we never touch a manifest that doesn't exist at a registry.
 
               if [[ ! -f /tmp/.buildx_metadata.json ]]; then
```
marians: **[cleanup]** This metadata-guard + index-digest extraction block is byte-identical to lines 333–342 in the cosign-refs step. The digest is content-addressed and shared — computing it once and persisting it (like `/tmp/.image_access` at line 310) keeps the two copies from drifting.

src/commands/image-build-and-push.yaml
```diff
@@ -416,20 +457,69 @@ steps:
...
+                    attest_digest=$(jq -r --arg d "$platform_digest" '
+                      .manifests[]
+                      | select(.annotations["vnd.docker.reference.type"] == "attestation-manifest"
+                               and .annotations["vnd.docker.reference.digest"] == $d)
+                      | .digest' /tmp/.index.json | head -n1)
+                    if [[ -z "$attest_digest" ]]; then
+                      echo "ERROR: no buildx attestation manifest found for $plat_ref (sbom=true expects an SPDX attestation)." >&2
+                      exit 1
+                    fi
+                    docker buildx imagetools inspect "${base}@${attest_digest}" --raw > /tmp/.attest_manifest.json
```
marians: **[efficiency]** The attestation manifest and SPDX blob are re-fetched once per *registry*, but they're content-addressed and identical across registries (same digests) — only the signing genuinely needs to be per-registry ref. Fetching once per platform (keyed by `$platform_digest`) and reusing the file saves (N−1)×P registry round-trips per build on a path that now runs on every default public build.

src/commands/image-build-and-push.yaml
```diff
@@ -354,19 +361,33 @@ steps:
               cat /tmp/.cosign_refs
         - cosign-sign-verify
   - when:
-      condition: <<parameters.sbom-cyclonedx>>
+      # Runs when there's any per-arch SBOM work to do: a CycloneDX SBOM was
+      # requested, or SPDX SBOMs are produced (sbom) and signing is on (sign).
+      condition:
```
marians, line 368: **[simplification]** Three layers now encode "is there SBOM work": this `when:`, the runtime early-exit (line 433), and the per-iteration guard (line 462) — and only the runtime one can be authoritative, since it alone sees `IMAGE_ACCESS`. The `sign` re-test in the line-462 guard is already folded into `SIGN_SBOMS`. Keeping the `when:` as a cheap skip is fine, but trimming the redundant re-tests (and noting that the bash check is the real gate) would keep the three layers from drifting.

src/commands/image-build-and-push.yaml
```diff
@@ -416,20 +457,69 @@ steps:
...
+                    fi
+                  fi
                 done < <(jq -r '
                   .manifests[]
                   | select(.platform.os != "unknown" and .platform.architecture != "unknown")
                   | "\(.digest) \(.platform.os)/\(.platform.architecture)\(if .platform.variant then "/"+.platform.variant else "" end)"
                 ' /tmp/.index.json)
               done < .registries_data
+
+              echo "Staged SBOM attestations to sign:"
```
marians, line 534: **[guideline]** Per the repo guideline (steps small, ideally one binary call; inter-step state cleaned up in a final cleanup step), this ~145-line run step leaves `/tmp/.cosign_attest`, `/tmp/sbom.*.json`, `/tmp/.index.json`, and `/tmp/.attest_manifest.json` behind with no cleanup — and `/tmp/.attest_manifest.json` is single-use and could be a pipe (`docker buildx imagetools inspect ... --raw | jq ...`). Consider splitting SPDX-staging and CycloneDX-staging into small commands mirroring the existing stage-refs / cosign-sign-verify decomposition, plus a cleanup step.

docs/cosign-signing.md
```diff
@@ -101,6 +111,32 @@ cosign verify-blob \
...
+  --type cyclonedx \
+  --certificate-oidc-issuer-regexp '^https://oidc\.circleci\.com' \
+  --certificate-identity-regexp '^https://circleci\.com/api/v2/projects/[a-f0-9-]+/pipeline-definitions/[a-f0-9-]+$' \
+  gsoci.azurecr.io/giantswarm/myapp@sha256:<platform-digest>
+```
+
+`verify-attestation` succeeds only if a matching signature exists and the
+signing identity matches; a substituted or tampered SBOM fails. To read the
+verified SBOM payload, pipe the output through
+`jq -r '.payload | @base64d | fromjson | .predicate'`.
```
marians, line 140: **[docs nit]** `cosign verify-attestation` prints one DSSE envelope per matching attestation, and attestations accumulate on re-runs of the same content-addressed digest — so this jq can emit multiple concatenated SPDX documents. Worth a sentence: "if the image was attested more than once you'll get one predicate per envelope line; pick the one you want (e.g. append `| jq -s 'last'`)".

src/commands/image-build-and-push.yaml
marians: I'm aware that `gitsemver validate --type dev` is the canonical method. Claude wasn't.

piontec:
Thanks for the thorough review @marians @QuentinBisson — addressed in ce689fd. Summary:

**Fixed**
- **`jq -e '.predicate'`** (both of you) — null/absent predicate now fails the pipeline under `pipefail` instead of signing a literal `null`.
- **Dev-tag divergence bug** — unified on the canonical `gitsemver validate --type dev`. Went further than a one-loop fix: the build step now computes the eligible registry set **once** into `/tmp/.eligible_registries`; the cosign and SBOM steps consume it, so the three drifting filter copies are gone (also closes the two `[cleanup]` comments on duplicated filter scaffold).
- **Duplicate index-digest block** — computed once into `/tmp/.index_digest`, consumed by both downstream steps.
- **`[efficiency]` per-registry refetch** — SPDX predicate and CycloneDX SBOM are now fetched/generated **once per platform** (content-addressed, identical across registries); only the per-registry ref staging / oras-attach remains per-registry.
- **`[design]` hard-fail after publish** — missing SPDX attestation manifest/layer is now **warn-and-skip**, since the image + index are already pushed and signed by that point.
- **`[ordering]`** — no-op exit moved ahead of the tool/metadata checks; `syft` required only when `sbom-cyclonedx: true` (`oras` always).
- **`[guideline]` cleanup** — added a final `when: always` step that removes the scratch files (`/tmp/.cosign_*`, `/tmp/sbom.*`, `/tmp/.index*`, `/tmp/.eligible_registries`, etc.).
- **`[fragility]`** — added a comment in `cosign-sign-verify` documenting the large-SBOM dependency on cosign's new-format (v0.3) sigstore bundles.
- **"byte-identical"** — reworded to "semantically identical / re-serialized" in the command comment, `docs/cosign-signing.md`, and `CHANGELOG.md`.
- **CHANGELOG `### Added` heading** in the released 9.0.2 section — restored (also restored the `cosign-sign-verify` `attest`-kind entry the main-merge had dropped).
- **`[behavior change]` public CycloneDX** — kept the signed attestation as the single source (per maintainer decision) and added an explicit **Changed** CHANGELOG note: public CycloneDX is now a signed sigstore-bundle referrer, not the `application/vnd.cyclonedx+json` referrer v9.0.2 documented; verify with `cosign verify-attestation --type cyclonedx`.
- **`[docs nit]`** — noted that `verify-attestation` can emit multiple DSSE envelopes (append `| jq -s 'last'`).

**Partially addressed — flagging for your call**
- **`[guideline]` split into separate commands** — I did the structural dedup + cleanup but kept SPDX/CycloneDX staging as cohesive `run` steps inside `image-build-and-push` rather than extracting new command files, to limit blast radius on an untestable-locally path. Happy to extract them into dedicated commands in a follow-up if you'd prefer.
- **`[simplification]` three gating layers** — the per-iteration `sign` re-test is folded into `SIGN_SBOMS`; I kept the `when:` as the cheap compile-time skip (you noted that's fine) with the bash check as the authoritative runtime gate.

QuentinBisson (approved):

PR #820, opened 2026-06-10
fix(cosign-sign-verify): retry sign-blob on Rekor duplicate entry conflict

marians:
## What

`cosign sign-blob` in `cosign-sign-verify` (kind `blob`) fails with:

```
[POST /api/v1/log/entries][409] createLogEntryConflict {"code":409,"message":"an equivalent entry already exists in the transparency log ..."}
```

Example: https://app.circleci.com/pipelines/github/giantswarm/search-mcp/396/workflows/21503f6f-dca3-43c6-b980-8c902ce233b6/jobs/782

## Why

#814 fixed this for `oci` and `attest` but left `blob` unchanged on the assumption that sign-blob has no Rekor dedup path. That assumption was wrong: with cosign v3, the sigstore bundle embeds the transparency log entry, so the Rekor upload happens inside `sign-blob` itself. The 409 occurs when the Rekor client retries a POST that already succeeded server-side.

## Fix

The skip-on-duplicate approach from #814 cannot be reused here: on a 409 the bundle file is never written, and `verify-blob --bundle` (and the release-asset upload) needs it. Instead, `cosign sign-blob` is retried up to 3 times. Each invocation generates fresh ephemeral keys, so a retry produces a new, non-equivalent Rekor entry and succeeds. Any non-409 error still fails immediately, and a persistent 409 after 3 attempts still fails the job.

### DIFF
--- a/src/commands/cosign-sign-verify.yaml
+++ b/src/commands/cosign-sign-verify.yaml
@@ -65,13 +65,19 @@ steps:
         ISSUER_REGEX='^https://oidc\.circleci\.com'
         IDENTITY_REGEX='^https://circleci\.com/api/v2/projects/[a-f0-9-]+/pipeline-definitions/[a-f0-9-]+$'
 
+        # Rekor duplicate-entry rejection (HTTP 409). Matched with `grep -E`.
+        # The structured token (createLogEntryConflict) appears in sign-blob
+        # errors; the prose is all the `oci`/`attest` errors carry, so both
+        # alternatives are needed.
+        REKOR_DUP_PATTERN='createLogEntryConflict|an equivalent entry already exists'
+
         case "<<parameters.kind>>" in
           oci)
             while IFS= read -r ref; do
               [[ -z "$ref" || "$ref" =~ ^# ]] && continue
               echo "cosign sign $ref"
               if ! out=$(cosign sign --yes "$ref" 2>&1); then
-                if echo "$out" | grep -q "an equivalent entry already exists"; then
+                if echo "$out" | grep -Eq "$REKOR_DUP_PATTERN"; then
                   echo "Rekor: entry already exists for $ref, skipping sign."
                 else
                   echo "$out" >&2; exit 1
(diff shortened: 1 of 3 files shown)


src/commands/cosign-sign-verify.yaml
```diff
@@ -85,10 +85,27 @@ steps:
...
             while IFS=' ' read -r file bundle; do
               [[ -z "$file" || "$file" =~ ^# ]] && continue
               echo "cosign sign-blob $file → $bundle"
-              cosign sign-blob --yes --bundle "$bundle" "$file"
+              for attempt in 1 2 3; do
+                rm -f "$bundle"
+                if out=$(cosign sign-blob --yes --bundle "$bundle" "$file" 2>&1); then
+                  break
+                fi
+                if echo "$out" | grep -q "an equivalent entry already exists" && [[ "$attempt" -lt 3 ]]; then
```
marians: This adds a third copy of the literal Rekor pattern `an equivalent entry already exists` (also at lines 74 and 129 in the `oci`/`attest` branches). It is prose from cosign's error output, so a cosign upgrade that rewords the message silently disables the 409 handling in all three branches at once — and all three sites have to be updated together.

Consider hoisting it into a single variable next to `ISSUER_REGEX`/`IDENTITY_REGEX`, e.g. `REKOR_DUP_PATTERN='an equivalent entry already exists'`, and optionally matching the structured token from the same error (`createLogEntryConflict`) instead of the prose, which is less likely to be reworded.

marians: The retry limit `3` is encoded in two places: the loop list here and the `[[ "$attempt" -lt 3 ]]` condition below. They only work because they currently agree — if the loop is later bumped to `1 2 3 4` without the condition (or vice versa), the final 409 either exits one attempt early or, worse, falls through the loop to `verify-blob` with `$bundle` already deleted by the `rm -f`, failing with a misleading missing-bundle error instead of a signing error.

```suggestion
              attempts=3
              for attempt in $(seq 1 "$attempts"); do
                rm -f "$bundle"
                if out=$(cosign sign-blob --yes --bundle "$bundle" "$file" 2>&1); then
                  break
                fi
                if echo "$out" | grep -q "an equivalent entry already exists" && [[ "$attempt" -lt "$attempts" ]]; then
```

marians: Done in 3556745: hoisted into `REKOR_DUP_PATTERN` next to the issuer/identity regexes, used via `grep -E` in all three branches. It matches both `createLogEntryConflict` and the prose — the structured token only appears in the `sign-blob` error; the `oci`/`attest` errors ("uploading to rekor: an equivalent entry already exists…") don't carry it, so the prose alternative has to stay.

marians: Done in 3556745: applied the suggestion — `attempts=3` with `seq`, and the last-attempt condition compares against `$attempts`.

QuentinBisson (commented):
LGTM

QuentinBisson (approved):

PR #827, opened 2026-06-11
Revert `no_output_timeout` and add hint what to do instead

AndiDog:
See #811

### DIFF
--- a/src/commands/image-build-and-push.yaml
+++ b/src/commands/image-build-and-push.yaml
@@ -96,8 +96,13 @@ steps:
   - run:
       name: Build and push multi-arch image to all registries
 
-      # Cross-platform builds may be slower (e.g. `GOARCH=arm64 go build` from another architecture)
-      no_output_timeout: 20m
+      # If you get the error "Too long with no output (exceeded 10m0s): context deadline exceeded", don't consider
+      # bumping `no_output_timeout` here. The default of 10 minutes is already a lot.
+      # The multi-architecture build, for example in Go, is too slow in Docker, so instead of building in Docker,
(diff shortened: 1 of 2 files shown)


src/commands/image-build-and-push.yaml
```diff
@@ -96,8 +96,13 @@ steps:
   - run:
       name: Build and push multi-arch image to all registries
 
-      # Cross-platform builds may be slower (e.g. `GOARCH=arm64 go build` from another architecture)
-      no_output_timeout: 20m
+      # If you get the error "Too long with no output (exceeded 10m0s): context deadline exceeded", don't consider
+      # bumping `no_output_timeout` here. The default of 10 minutes is already a lot.
+      # The multi-architecture build, for example in Go, is too slow in Docker, so instead of building in Docker,
+      # depend on `architect/go-build` and use something like
+      # `COPY mybinary-${TARGETOS}-${TARGETARCH} /my/destination`.
```
QuentinBisson, line 103: Can you link a PR that does that in the comment ?

hervenicol:
Even with the smallest Dockerfile, 10min may not be enough.
For instance observability-operator takes 15min (https://gigantic.slack.com/archives/C01176DKNP4/p1781693353455559)

AndiDog:
> Even with the smallest Dockerfile, 10min may not be enough. For instance observability-operator takes 15min ([gigantic.slack.com/archives/C01176DKNP4/p1781693353455559](https://gigantic.slack.com/archives/C01176DKNP4/p1781693353455559))

Okay, since we have that problem elsewhere, I'll just leave the timeout. We'll notice if CI pipelines are slow or costs increase.

PR #838, opened 2026-06-18
feat(go-build): cache GOCACHE and add opt-in parallel arch builds

QuentinBisson:
### Persist the Go build cache (the main win, on by default)
`go-cache-restore`/`go-cache-save` only cached `/go/pkg/mod` (module *downloads*). `GOCACHE` (compiled stdlib + dependency objects) was thrown away every run, so each run recompiled everything from scratch, once per `GOOS/GOARCH`.

- Pin `GOCACHE=/go/cache/go-build` (same volume as the module cache, no `$HOME` assumption) so both `go test` and `go build` populate it.
- Persist it keyed on `go.sum`, mirroring the module cache. The arch-multiplied cost is recompiling stdlib + dependencies, which only change when `go.sum` changes, so this captures the dominant win while storing **one cache object per dependency set, not one per commit** (important since this is on by default across every consuming repo). The first run after a `go.sum` change is cold; later runs reuse compiled stdlib/deps and only recompile the repo's own changed packages.

### Opt-in parallel architecture builds
- New `build_concurrency` parameter: `"1"` (default, sequential), `"auto"` (= `nproc`), or an integer.
- Default stays sequential: a single `go build` already saturates a `medium` executor's 2 vCPUs, so parallelism only pays with a larger `resource_class`. Docs recommend pairing `build_concurrency` with `large`/`xlarge` for 3+ architectures, with the credit math.

### Failure-masking fixes (surfaced by backgrounding)
- `build_one` ended on the `linux/amd64` `if` test, returning `0` even when `go build` failed for a non-amd64 target. It now returns the build's real status. The previous code only avoided this via `set -e`, which is suppressed by both `&` and `|| fail=1`.
- Exit codes are aggregated by waiting each PID exactly once (batched waves), not via `jobs -rp` (which lists only *running* jobs and dropped the status of fast-failing builds).

## Measured impact (live on muster, CircleCI `medium` = 2 vCPU)

Ran the dev orb against `giantswarm/muster` (2 arches, `build_concurrency: "2"`): one cold run, then a warm run with `go.sum` unchanged.

| go-build step | Cold (today) | Warm | Δ |
|---|---|---|---|
| Restore Go build cache | 0.1s (miss) | 15.8s (hit) | — |
| Run unit tests | 115.6s | 17.8s | 6.5× |
| Build binaries | 194.7s | 5.0s | 39× |
| Save Go build cache | 23.9s | 0.1s (key exists) | — |
| **Total job** | **392.6s** | **76.4s** | **5.1×** |

The `Save Go build cache` dropping to 0.1s on the warm run confirms the write-once `go.sum` key: no per-commit re-upload, bounded storage.

## Cost estimate

`medium` = 10 credits/min, 1 credit = $0.0006 ($0.006/min).

- **Measured (2 arches):** 392.6s → 76.4s = ~5.3 min saved per warm run ≈ **53 credits (~$0.032)**.
- **Projected to muster's real 6-arch config** (build-binaries scales ~linearly per arch, ~97s/arch cold vs ~2.5s warm; other steps arch-independent): cold ~12 min → warm ~1.5 min ≈ **~10 min, ~100 credits (~$0.06) saved per warm run**.

Per-run dollars are small; the value compounds on **frequency × fleet**. Illustrative single busy Go repo (~80 warm go-build runs/month): **~$3–5/month and ~10+ hours of CI wall-clock saved**. Across the orb's installed base (~13,660 builds / 254 projects in 30 days), every Go repo's go-build job picks this up automatically with no config change.

**Caveats:** the first build after any `go.sum` change is still cold (e.g. renovate bumps), so not every run is warm — but cold was the *only* mode before. Offsetting cost is one extra cache object per dependency set (~1–2 GB at 6 arches), refreshed only when deps change. Parallelism contributes little at the default `medium`/sequential config; the cache carries essentially all of the win.

## Notes
- Cache is on by default (transparent, same artifacts); parallelism is opt-in (needs provisioned CPU). `build_concurrency: "1"` keeps existing behaviour, so no resource/credit change for current consumers unless they opt in.
- Existing contracts unchanged: ldflags injection, the bare `<binary>` linux/amd64 copy, `.platforms`, and cosign signing all stay as-is.

## Verification
- `circleci orb pack src/` + `circleci orb validate` both pass.
- Ran the reworked build loop against a real Go module: `build_concurrency` in {1, 2, auto} each produce all per-arch binaries + the bare copy + a clean 3-line `.platforms`; a deliberately broken arch returns non-zero in all three modes.
- End-to-end on `giantswarm/muster` via a dev publish of this branch: cold then warm run, cache restore hit confirmed, timings in the table above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Added
+
+- `go-build`: `build_concurrency` parameter to compile multiple architectures concurrently (`"1"` sequential by default, `"auto"`, or an integer).
+
+### Changed
+
+- `go-build`: persist the Go build cache (`GOCACHE`) across runs in addition to the module cache, so warm builds only recompile changed packages.
+
 ## [9.4.2] - 2026-06-17
 
 ### Changed
--- a/docs/job/go-build.md
+++ b/docs/job/go-build.md
@@ -75,12 +75,13 @@ See [`push-to-registries`](./push-to-registries.md) for the full Dockerfile cont
 
 ## Build speed and concurrency
 
-The job persists two caches across runs: the Go module cache (`/go/pkg/mod`,
-keyed on `go.sum`) and the Go build cache (`GOCACHE`, keyed on `go.sum` plus
-branch and revision with prefix fallbacks). The build cache holds compiled
-stdlib and dependency objects for every GOOS/GOARCH, so a warm cache only
-recompiles packages that actually changed. This is the largest single factor in
-build time; the first run after a `go.sum` change is cold and slower.
+The job persists two caches across runs, both keyed on `go.sum`: the Go module
+cache (`/go/pkg/mod`) and the Go build cache (`GOCACHE`). The build cache holds
+compiled stdlib and dependency objects for every GOOS/GOARCH, which are the bulk
+of build time and only change when `go.sum` changes; keying on `go.sum` keeps one
+cache object per dependency set rather than one per commit. The first run after a
+`go.sum` change is cold and slower; subsequent runs reuse the compiled stdlib and
+dependencies and only recompile the repo's own changed packages.
 
 Cross-compiling each architecture is independent work. `build_concurrency`
 controls how many run at once:
--- a/docs/job/go-build.md
+++ b/docs/job/go-build.md
@@ -16,6 +16,7 @@ Builds Go binaries for one or more target architectures in a single job and pers
 - `pre_test_target`: Makefile target to run before tests/lints (optional).
 - `tags`: Additional Go build tags (optional).
 - `test_target`: Makefile target to run for tests (optional).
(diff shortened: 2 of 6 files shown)


teemow (approved):
Approving. Reviewed the full diff.

- `GOCACHE` pinned to `/go/cache/go-build` via `BASH_ENV` in `go-cache-restore` (runs before go-test/go-build), saved + restored on a `go.sum`-keyed cache parallel to the module cache. `go.sum`-only key is the right call: `save_cache` is a no-op when the key exists, so it's write-once per dependency set (warm save ~0.1s), bounded storage; changed repo packages still recompile via Go's content-addressed cache.
- `build_concurrency` defaults to `"1"` (sequential), so no behaviour/credit change for current consumers; parallel waves wait each PID once and aggregate exit status; `.platforms` is written serially.
- Bonus: `build_one` now returns the real `go build` status (`if ! ... then return 1`), closing a failure-masking hole where a non-amd64 build failure could return 0 under `&` / `|| fail=1`.
- Docs are thorough (incl. the resource_class/credit math).

Measured 5.1x warm on muster (392.6s -> 76.4s). LGTM.

Gacko:
Please have a look at this: https://app.circleci.com/pipelines/github/giantswarm/clustertest/2660/workflows/62b4969d-2016-4c47-9a59-dedc73e6a82d/jobs/2670. I assume your change is somehow connected. 🙂

PR #839, opened 2026-06-18
feat(run-tests-with-ats): run app-test-suite container directly, drop dats.sh

QuentinBisson:
## What

`run-tests-with-ats` no longer downloads and runs the `dats.sh` wrapper from the `app-test-suite` repo. It runs the `app-test-suite` container image directly with `docker run`.

## Why

`app-test-suite` has removed `dats.sh` (giantswarm/app-test-suite#642). The command currently does `wget https://raw.githubusercontent.com/giantswarm/app-test-suite/<version>/dats.sh`, which now 404s for any consumer pinning to a current ref. `dats.sh` was only a thin `docker run` wrapper, so the orb can invoke the container itself.

## Changes

- `src/commands/run-tests-with-ats.yaml`: replace the `Download dats.sh` + `./dats.sh` steps with a direct `docker run gsoci.azurecr.io/giantswarm/app-test-suite:<tag> --chart-file ... <flags>` (same uid/gid, docker.sock, and `--network host` the wrapper used; `-it` dropped for CI).
- Rename parameter `app-test-suite_container_tag` to `app-test-suite_version` (it now selects the container image tag; previously `app-test-suite_version` selected the `dats.sh` git ref). Default `1.0.0`.
- Update `docs/job/run-tests-with-ats.md`.

## Migration (BREAKING, requires major → v10.0.0)

- Consumers setting `app-test-suite_container_tag:` must rename it to `app-test-suite_version:`.
- Consumers who previously set `app-test-suite_version:` to a `dats.sh` git ref must change the value to a container image tag (e.g. `1.0.0`).
- `giantswarm/architect@9` keeps the old `dats.sh` path; bump to `@10` for direct container execution.

Cut the release via the `main#release#major` trigger. CircleCI orb pack/validate runs in CI (no local `circleci` CLI).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Changed
+
+- **BREAKING:** `run-tests-with-ats` runs the `app-test-suite` container directly with `docker run` instead of
+  downloading and executing the `dats.sh` wrapper, which has been removed from the `app-test-suite` repository.
+  The `app-test-suite_version` parameter is removed (it only selected the `dats.sh` ref); use
+  `app-test-suite_container_tag` to pick the image. Requires a new major version.
+
 ## [9.4.2] - 2026-06-17
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - **BREAKING:** `run-tests-with-ats` runs the `app-test-suite` container directly with `docker run` instead of
   downloading and executing the `dats.sh` wrapper, which has been removed from the `app-test-suite` repository.
-  The `app-test-suite_version` parameter is removed (it only selected the `dats.sh` ref); use
-  `app-test-suite_container_tag` to pick the image. Requires a new major version.
+  The `app-test-suite_container_tag` parameter is renamed to `app-test-suite_version` and now selects the
+  container image tag (previously it selected the `dats.sh` git ref). Its default is `"1.0.0"`.
+  Requires a new major version.
 
 ## [9.4.2] - 2026-06-17
 
--- a/docs/job/run-tests-with-ats.md
+++ b/docs/job/run-tests-with-ats.md
@@ -35,7 +35,7 @@ workflows:
 
 - [common parameters](common.md#parameters) shared in all jobs.
 - [chart_archive_prefix](#chart_archive_prefix-optional-string-default)
-- [app-test-suite_container_tag](#app-test-suite_container_tag)
+- [app-test-suite_version](#app-test-suite_version)
 - [additional_app-test-suite_flags](#additional_app-test-suite_flags)
 
 ### chart_archive_prefix (optional string, default="")
@@ -48,12 +48,12 @@ Actual implementation is:
 find -name "<< parameters.chart_archive_prefix >>*.tgz" -print -quit
 ```
 
-### app-test-suite_container_tag
+### app-test-suite_version
 
-Container tag of app-test-suite to use (check gsoci.azurecr.io/giantswarm/app-test-suite).
+app-test-suite container image tag to run (check gsoci.azurecr.io/giantswarm/app-test-suite).
 The job runs this container image directly with `docker run`.
 
-(Default: "0.15.0")
+(Default: "1.0.0")
 
 ### additional_app-test-suite_flags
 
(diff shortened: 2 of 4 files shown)


teemow:
Thanks for taking `dats.sh` out of the loop — Bumblebee needs exactly this to get `agent-platform-standalone` onto app-test-suite `main`. Two things I ran into while trying to use it against `main` (`0.15.1-dev…h4162ff7` dev image), plus a proposal:

**Gap.** Since giantswarm/app-test-suite#675 ats no longer creates a cluster: `--cluster-kubeconfig` is required, and `kind`/`docker` left the image. The `docker run` here passes neither a kubeconfig nor creates a cluster, so against `main` the job fails at config validation ("Path to the cluster 'kubeconfig' file must be configured"). The Docker-socket mount and `DOCKER_GID` are no longer needed on that path.

**Rename.** Twelve repos set `app-test-suite_version` / `app-test-suite_container_tag` explicitly (cert-manager-app, rbac-operator, cert-exporter, metrics-server-app, app-exporter, flux-app, crossplane, external-secrets, sloth-app, konfigure-operator, dex-app, kyverno-app). Dropping `app-test-suite_container_tag` fails their config compilation, and `app-test-suite_version` silently changes meaning for the ones that set it to a git ref. Since v10 already shipped (2026-08-18), this would now be v11.

**Proposal** (branch [`run-tests-with-ats-kind`](https://github.com/giantswarm/architect-orb/compare/main...run-tests-with-ats-kind), published as `giantswarm/architect@dev:run-tests-with-ats-kind`; I have not opened a PR so as not to compete with this one):

- `run-tests-with-ats` gains `create_kind_cluster` (boolean, default `false`), `kind_version` and `kind_node_image` (default the `gsoci.azurecr.io/giantswarm/kind-node` mirror the `integration-test` job already uses, so no Docker Hub pulls). With `create_kind_cluster: true` the job installs kind, creates a cluster with the kubeconfig written to `kube.config` in the workdir, runs the image directly with `--cluster-kubeconfig kube.config --cluster-type kind --cluster-version <from the node image tag>` (no socket), and deletes the cluster in a `when: always` step.
- Both existing parameters keep their names and defaults; a job that does not set `create_kind_cluster` runs the `dats.sh` path byte-for-byte. That makes it a **minor** release, and the fleet can migrate repo by repo — the migration also touches repo-owned files (`.ats/main.yaml` loses the `*-cluster-type` keys, `tests/ats/Pipfile` becomes `pyproject.toml` + `uv.lock`, `app_factory` tests install with Helm), so a default flip cannot be a flag day. The default flip is then a later major, once the consumers have moved.
- Renovate regex managers for the two new pins; docs and CHANGELOG entry included.

The consumer side: `agent-platform-standalone` is being migrated on the dev orb + the `main` dev image (I will link the green run here), devctl's generated CI gets an `--ats-version` knob that emits the two job parameters, and giantswarm/app-test-suite#713 asks for the 1.0.0 release itself (with the PyPI trusted-publisher gap I found).

Happy to fold the kind provisioning into this PR as a follow-up commit on your branch if you prefer that over the separate branch — your call and Honeybadger's on minor vs. major. If I don't hear back in a few days I'll open the branch as a PR so it can be reviewed on its own.

teemow:
Evidence for the branch: giantswarm/agent-platform-standalone#140 ran its chart smoke on `giantswarm/architect@dev:run-tests-with-ats-kind` with `create_kind_cluster: true` and the app-test-suite `main` dev image — [job 1595](https://app.circleci.com/pipelines/github/giantswarm/agent-platform-standalone/jobs/1595): green, 226 s end to end (install kind 0.6 s, create cluster 38.9 s, app-test-suite 182 s of which pytest 157 s with 5 passed, delete cluster 2.9 s) against a 237 s baseline on the `dats.sh` path with apptestctl. The generated-CI side is giantswarm/devctl#2152 (`--ats-version`, draft until an orb release carries the parameter) and giantswarm/github#5886.

teemow:
Timo asked Bumblebee to drive this through, so the kind-provisioning variant is now open as #917 (same parameters kept, `create_kind_cluster` opt-in, shippable as a minor), with the release side in giantswarm/app-test-suite#713. If #917 lands, this one can be closed — your call, I will not touch it.

teemow:
Shipped: #917 (`create_kind_cluster` on `run-tests-with-ats`, opt-in, both legacy params and the dats.sh path kept) merged and released as architect-orb v10.3.0 today. devctl pins it and gains `gen circleci --ats-version` (giantswarm/devctl#2152); app-test-suite releases 1.0.0 via giantswarm/app-test-suite#714. The default flip to the direct path is a later major, Honeybadger's call. @QuentinBisson this supersedes the change here, so feel free to close this PR.

PR #872, opened 2026-08-06
chore: remove ct_config and skip_conftest_deprek8ion params from push-to-app-catalog

weatherhog:
Retires the two `push-to-app-catalog` parameters that the orb has ignored for at least a major version. **Breaking — targets v10.0.0.**

| Param | Ignored since | Repos still passing it |
|---|---|---|
| `skip_conftest_deprek8ion` | v6.3.2 (`helm-conftest` step removed) | 2 |
| `ct_config` | v9.0.0 (`helm-lint` step removed) | 20 |

### Consumers are already cleaned

All 22 consumer repos have PRs open removing the params, so this can be released without breaking anyone. Tracking list is in giantswarm/roadmap#4066.

### `executor` is deliberately *not* removed

v9.0.0 parked three compat params for "a future version". Only two go here. `executor` accepts a single valid value that is also the default, so passing it is already a no-op — but **~275 repos set it explicitly**, and removing it would break all of them for zero benefit. Its description is updated to say it is permanently retained, so it stops reading as pending removal.

### Also in this PR

- `docs/migration-v9-to-v10.md`, matching the v8→v9 precedent, so the Renovate major PRs have something to link to. It makes the point that **`ct-config.yaml` files do not need to change** — app-build-suite still reads chart-testing config via `ct-config` in `.abs/main.yaml`; only the orb parameter is gone.

### Release

**Merging this does not release it.** Production publish is tag-driven, and the tag comes from merging a release PR created by pushing `main#release#major`. That should happen only once the 22 consumer PRs have actually merged.

Companion image PRs: giantswarm/architect#1391, giantswarm/app-build-suite#581, giantswarm/app-test-suite#699.

Part of giantswarm/roadmap#4066

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,23 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Changed
 
 - `push-to-registries`: register QEMU/binfmt handlers only when at least one target platform differs from the build host. Single-arch repos (`platforms: linux/amd64`) were paying a privileged `tonistiigi/binfmt --install all` image pull on every build for handlers that could never be used, since no `RUN` step is emulated when the target matches the host. Multi-arch builds are unchanged, and if the host platform cannot be determined the handlers are registered as before rather than risking an un-emulated cross build.
+- `executor` parameter on `push-to-app-catalog` is now **permanently retained** rather than pending removal.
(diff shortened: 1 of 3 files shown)


mproffitt (approved):
@weatherhog Resolve merge conflicts and then this looks good.

PR #880, opened 2026-08-17
refactor: extracts registry selection and image signing into commands

JosephSalisbury:


src/commands/image-build-and-push.yaml
```diff
@@ -133,6 +133,13 @@ steps:
       condition: <<parameters.push>>
       steps:
         - generate-github-token
+        # Resolves the push set into /tmp/.eligible_registries, /tmp/.image_access
+        # and /tmp/.cache_registry. Build-only mode skips it: nothing leaves the
+        # BuildKit cache, so there is no set to select and no credentials to need.
+        - image-select-registries:
```
JosephSalisbury, line 139: Everything the old file did at lines 197–307 — the visibility check, the registries-data load, and the eligible-set + cache-host loop — now happens in this one call. Moved to `image-select-registries.yaml:57–151`.

The `when: <<parameters.push>>` gate around it preserves the old behaviour exactly: with `push: false` the baseline never reached that code either, because it sat inside the `else` branch of the push check.

src/commands/image-build-and-push.yaml
```diff
@@ -194,117 +201,32 @@ steps:
...
-                echo "Tag the image for $reg"
-                echo "Preparing tag for $reg: <<parameters.image>>:${tag##*:}"
-                all_tags+=("--tag" "$reg/<<parameters.image>>:${tag##*:}")
-              done
-            else
-              echo "Registry $reg is not configured for $IMAGE_ACCESS images, skipping"
-            fi
-          done < .registries_data
-          cache_registry="${cache_registry:-$cache_fallback}"
+          while IFS= read -r reg; do
```
JosephSalisbury, line 221: This is the one block that deliberately did **not** move, and the only logic change in this file.

- **Baseline** (lines 264–307): built `--tag` flags *inside* the loop that filtered `.registries_data`, so filtering and tag-building were interleaved.
- **Now**: the filtering lives in `image-select-registries`, and this reads the resolved set back from `/tmp/.eligible_registries` and does a plain cross-product with `tags`.

Same output. It stayed here rather than moving because the per-architecture legs in the follow-up PR push by digest with no tags at all, so tag-building is specific to this path.

src/commands/image-build-and-push.yaml
```diff
@@ -508,233 +432,13 @@ steps:
...
-              /tmp/.index.json /tmp/.attest_manifest.json \
-              /tmp/.index_digest /tmp/.eligible_registries \
-              /tmp/.image_access /tmp/.buildx_metadata.json \
-              /tmp/.gh_response /tmp/.buildx_output \
-              /tmp/sbom.*.spdx.json /tmp/sbom.*.cdx.json 2>/dev/null || true
+  # Signs the index, signs or attaches the per-architecture SBOMs, and clears the
+  # inter-step scratch files. Invoked unconditionally: the signing work carries
+  # its own conditions, and the cleanup inside runs `when: always`, so build-only
+  # and failed runs still leave nothing behind for an unrelated rerun.
+  - image-sign-and-attest:
```
JosephSalisbury, line 439: Baseline lines 511–741 — both `when:` blocks (cosign refs + `cosign-sign-verify`, then per-arch SBOM staging + attest) and the `when: always` cleanup — moved to `image-sign-and-attest.yaml:51–281`. 231 lines, 228 of them byte-identical.

Called unconditionally on purpose: the two signing blocks kept their own `when:` conditions, and the cleanup is `when: always`, so a `push: false` or failed run still cleans up exactly as before.

src/commands/image-select-registries.yaml
```diff
@@ -0,0 +1,160 @@
...
+    default: false
+steps:
+  - run:
+      name: "architect/image-select-registries: Resolve the eligible registry set"
+      command: |
+        set -o pipefail
+
+        TAG_VALUE="${DOCKER_IMAGE_TAG}"
+
+        # If image is forced to be pushed as a public one,
```
JosephSalisbury, line 57: From here to line 151 is baseline `image-build-and-push.yaml:197–307`. 90 of these 95 lines are byte-identical; the 5 that differ are the reworded comment at 108–112, which used to describe the tag-building this command no longer does.

To confirm:

```sh
diff -w <(git show main:src/commands/image-build-and-push.yaml | sed -n '197,307p' | sed 's/^[[:space:]]*//') \
        <(sed -n '57,151p' src/commands/image-select-registries.yaml | sed 's/^[[:space:]]*//')
```

Expected: the tag-building lines removed, and that one comment reworded. Nothing else.

src/commands/image-select-registries.yaml
```diff
@@ -0,0 +1,160 @@
...
+          if [[ "$access" == *"$IMAGE_ACCESS"* ]]; then
+            echo -n ${reg} > /tmp/.docker_image_registry
+            echo "$reg" >> /tmp/.eligible_registries
+          else
+            echo "Registry $reg is not configured for $IMAGE_ACCESS images, skipping"
+          fi
+        done < .registries_data
+        cache_registry="${cache_registry:-$cache_fallback}"
+
+        echo -n "$IMAGE_ACCESS" > /tmp/.image_access
```
JosephSalisbury, line 153: These two lines are the new logic in this PR.

`IMAGE_ACCESS` and `cache_registry` used to be shell variables inside a single `run:` step, so they needed no serialising. They now cross a command boundary, so they go to files — the same pattern `.eligible_registries` already used.

One consequence worth a look while you are here: `/tmp/.image_access` is now written **before** the build rather than after it (the baseline wrote it at line 499, post-push). Nothing reads it earlier and `image-sign-and-attest` only runs after the build, so the reordering is inert — but it does mean a failed build now leaves the file behind, which the `when: always` cleanup removes.

src/commands/image-sign-and-attest.yaml
```diff
@@ -0,0 +1,281 @@
...
+    type: boolean
+    default: true
+  sbom-cyclonedx:
+    description: |
+      Additionally generate a CycloneDX SBOM per architecture (with syft), signed
+      on public images and attached as an unsigned oras referrer otherwise.
+    type: boolean
+    default: false
+steps:
+  - when:
```
JosephSalisbury, line 51: From here to line 281 is baseline `image-build-and-push.yaml:511–741`, unchanged apart from line 278.

```sh
diff -w <(git show main:src/commands/image-build-and-push.yaml | sed -n '511,741p' | sed 's/^[[:space:]]*//') \
        <(sed -n '51,281p' src/commands/image-sign-and-attest.yaml | sed 's/^[[:space:]]*//')
```

Expected: one line replaced by two — `/tmp/.cache_registry` added to the `rm -f`, because that file is new inter-step state.

JosephSalisbury (commented):
Review aid, not a change request — the diff hides that this is almost entirely a move.

Accounting for the 441 lines across the two new files:

| Category | Lines |
| --- | --- |
| Moved verbatim from `image-build-and-push` | 326 |
| New: header comments | 51 |
| New: parameter declarations | 46 |
| New: `steps:`/`run:` scaffolding | 9 |
| **New: actual logic** | **9** |

Those 9: two `echo > /tmp/...` writes, five lines logging them, plus one extra path in the cleanup `rm -f`.

The inline comments below mark each boundary and give a `diff` to confirm it.

JosephSalisbury:
Split into #882 and #883, per the two halves being very different to review.

The signing lift (#882) is 230 lines with **zero** deltas — the moved code was already communicating with the build step purely through files, so it lifts byte-identically and the verification `diff` prints nothing at all.

The registry-selection lift (#883) is +191/-106 and contains all of the actual thinking that was buried in this PR: the tag-building relocation, the shell-state-to-files conversion, and the `/tmp/.image_access` write moving from post-build to pre-build.

Combining them meant those ~10 interesting lines sat inside a 481-line diff with no way for a reviewer to know that 230 of them were a no-brainer.

PR #882, opened 2026-08-17
refactor: extracts image signing and SBOM steps into a command

JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/37408


JosephSalisbury (commented):
Review aid: `image-sign-and-attest.yaml` is a new file but not new code. 231 of its 279 lines are `image-build-and-push.yaml:511–741` lifted out; the other 48 are the header comment and the parameter block.

Two functional changes in the moved region, both marked inline below:

- **line 150** — a log message that named the build step as the producer of the eligible registry set
- **lines 276–277** — `/tmp/.cache_registry` added to the cleanup `rm -f`

Everything else that differs is comment rewording.

```sh
# the moved region: prints the two changes above, plus comment rewording
diff -w <(git show main:src/commands/image-build-and-push.yaml | sed -n '511,741p' | sed 's/^[[:space:]]*//') \
        <(sed -n '49,279p' src/commands/image-sign-and-attest.yaml | sed 's/^[[:space:]]*//')

# the caller below the extraction point: prints nothing
diff <(git show main:src/commands/image-build-and-push.yaml | sed -n '1,510p') \
     <(sed -n '1,510p' src/commands/image-build-and-push.yaml)
```

src/commands/image-sign-and-attest.yaml
```diff
@@ -0,0 +1,279 @@
...
+    type: boolean
+    default: true
+  sbom-cyclonedx:
+    description: |
+      Additionally generate a CycloneDX SBOM per architecture (with syft), signed
+      on public images and attached as an unsigned oras referrer otherwise.
+    type: boolean
+    default: false
+steps:
+  - when:
```
JosephSalisbury: Everything from here to line 279 is `image-build-and-push.yaml:511–741` on `main` — both `when:` blocks (cosign refs, then per-arch SBOM staging) and the `when: always` cleanup.

The lift is possible because these steps only ever talked to the build step through files — `/tmp/.image_access`, `/tmp/.eligible_registries`, `/tmp/.index_digest` — and never knew how the index was produced. That is also what makes them correct against an index assembled by `imagetools create`, which is what the top of this stack needs.

src/commands/image-sign-and-attest.yaml
```diff
@@ -0,0 +1,279 @@
...
+              # SBOM work) needs neither syft/oras nor the index digest.
+              if [[ "$SIGN_SBOMS" != "true" && "<<parameters.sbom-cyclonedx>>" != "true" ]]; then
+                echo "No SBOM work: signing skipped (private or sign=false) and CycloneDX not requested."
+                exit 0
+              fi
+
+              # Eligible registries and index digest, both resolved before this command runs.
+              mapfile -t ELIGIBLE_REGISTRIES < /tmp/.eligible_registries
+              if [[ ${#ELIGIBLE_REGISTRIES[@]} -eq 0 ]]; then
+                echo "No eligible registries recorded; nothing to do."
```
JosephSalisbury: The only non-comment change in the moved region. Was:

```
echo "No eligible registries recorded by the build step; nothing to do."
```

User-visible output, and the next PR in the stack moves that derivation into `image-select-registries`, so naming the build step would print the wrong producer.

src/commands/image-sign-and-attest.yaml
```diff
@@ -0,0 +1,279 @@
...
+      # failure so a failed build doesn't leave stale predicates/digests behind
+      # for an unrelated rerun on the same runner. The split-china-push cache uses
+      # /tmp/.docker_image_* (not touched here), so it is unaffected.
+      when: always
+      command: |
+        rm -f /tmp/.cosign_refs /tmp/.cosign_attest \
+              /tmp/.index.json /tmp/.attest_manifest.json \
+              /tmp/.index_digest /tmp/.eligible_registries \
+              /tmp/.image_access /tmp/.cache_registry \
+              /tmp/.buildx_metadata.json \
```
JosephSalisbury: `/tmp/.cache_registry` added to the cleanup. Nothing writes that file yet — it arrives with `image-select-registries` in the next PR — and this is an `rm -f`, so it is inert here.

It is in this PR rather than the next one so this command owns the whole lifecycle of the scratch state it cleans up, and the follow-up does not have to reach back into a file it otherwise has no business editing.

src/commands/image-build-and-push.yaml
```diff
@@ -508,233 +508,13 @@ steps:
...
-              /tmp/.index.json /tmp/.attest_manifest.json \
-              /tmp/.index_digest /tmp/.eligible_registries \
-              /tmp/.image_access /tmp/.buildx_metadata.json \
-              /tmp/.gh_response /tmp/.buildx_output \
-              /tmp/sbom.*.spdx.json /tmp/sbom.*.cdx.json 2>/dev/null || true
+  # Signs the index, signs or attaches the per-architecture SBOMs, and clears the
+  # inter-step scratch files. Invoked unconditionally: the signing work carries
+  # its own conditions, and the cleanup inside runs `when: always`, so build-only
+  # and failed runs still leave nothing behind for an unrelated rerun.
+  - image-sign-and-attest:
```
JosephSalisbury: 231 lines become these 10.

Everything above this point is byte-identical to `main`, which is the other half of "nothing else changed":

```sh
diff <(git show main:src/commands/image-build-and-push.yaml | sed -n '1,510p') \
     <(sed -n '1,510p' src/commands/image-build-and-push.yaml)
```

Invoked unconditionally, deliberately: the two signing blocks kept their own `when:` conditions and the cleanup kept `when: always`, so `push: false` and failed runs behave exactly as before. Gating the call here would change that.

JosephSalisbury:
Correction to the summary above: the `/tmp/.cache_registry` cleanup line has moved to the next PR in the stack, where the file it removes is introduced. Having it here meant this PR carried a line whose only purpose was to serve the next one — a forward reference to state that does not exist yet.

So the moved region now differs from the original in three places, all text: two comments and one log line stop naming "the build step" as the producer of the eligible registry set. No logic changed, and nothing here references anything that has not landed.

PR #897, opened 2026-08-19
feat(push-to-app-catalog): report published chart version on the PR

mcharriere:
## What

`architect/push-to-app-catalog` always stamps the chart version from `gitsemver`. Off-tag that yields `X.Y.Z-dev.<branch>.<date>.<time>` — collision-proof, but not guessable, so a reviewer who wants to install a PR's chart has to open the CircleCI job and read the log.

This adds a new `comment-chart-version` command that posts a single sticky comment on the pull request with the chart name, version, OCI reference, digest, a `helm pull` command and a link to the git catalog. Later pushes update that same comment in place, so a long-running branch produces one comment and one notification rather than one per push.

Enabled by default via a new `comment_on_pr` parameter on the job.

## How

- **New** `src/commands/comment-chart-version.yaml`. Sticky comment found by a hidden marker keyed on the `chart` parameter, so a repo publishing several charts gets one comment per chart and the key survives a `Chart.yaml` rename mid-PR. `PATCH` when found, `POST` otherwise.
- **PR resolution** goes through `GET /repos/{o}/{r}/commits/{sha}/pulls`, not `CIRCLE_PULL_REQUEST` — CircleCI does not provide that under its GitHub App integration, and it is unreliable for same-repo PRs under OAuth. This repo's own config already sets `enable_pr_comment: false` on `orb-tools/publish` for that reason. The commit-keyed endpoint also covers forked PRs and matches a PR opened *after* the push. `?head=owner:branch` remains as a fallback.
- `package-helm-with-abs` records `.build_chart_version` and derives `.build_chart_name` from the produced `build/*.tgz`. Reading it off the artifact rather than the `chart` parameter keeps it correct under `explicit_allow_chart_name_mismatch`, mirroring why `push-helm` already derives the pushed ref from `helm push` output.
- `push-helm` persists `/tmp/.build_chart_oci_ref` and `/tmp/.build_chart_digest`. Kept separate from `/tmp/.cosign_refs.candidate`, which is digest-form with no tag and whose lifecycle belongs to signing.
- `generate-github-token` gains `non_fatal` (default `false`, so the four existing call sites are untouched), and now checks `gh-token`'s exit status — previously a failure left the step green with an empty token, surfacing later as a confusing "GITHUB_TOKEN is not set".
- Table rows are omitted, not blanked, when the data is absent, driven off file existence rather than the parameters — so a row only appears if that push actually happened.

## Non-fatal by construction

CircleCI has no per-step `continue-on-error`, so the script turns off `-e` and runs its logic in a subshell with a trailing bare `exit 0`. Nothing is posted, and the build stays green, for: tag builds, default-branch builds, forked PRs (no context, so no token), a PR that does not exist yet, a missing App permission, or a rate limit. The chart has already been published by the time this runs.

## Prerequisite — needs action before release

The `CircleCI architect` GitHub App currently only needs *Metadata: read*. Posting a comment needs **`Pull requests: write`** (plus `Issues: write`, since PR conversation comments live on the issues endpoints), and an organisation owner has to accept the updated permission set on the installation.

Because `comment_on_pr` defaults to `true`, until that is granted every chart publish will log a one-line warning naming the missing permission. Builds stay green, but the grant should land before this is released.

## Testing

The orb can only really be exercised from a consumer repo, so the shell was extracted from the YAML and run against a mocked GitHub API first:

- 13 scenarios: create; in-place `PATCH`; `PATCH` 404 → `POST` fallback; 403 permission failure; no PR; `?head=` fallback; tag build; default branch; no token; chart facts not recorded; private chart; `push_to_oci_registry: false`; published name ≠ `chart` param; two open PRs. All exit 0.
- Pagination: marker on page 2 of a >100-comment PR is found and patched; hitting the 10-page cap logs a warning rather than truncating silently.
- 8 `generate-github-token` paths: missing key, `gh-token` failure and empty output, each with `non_fatal` true and false, plus the pre-minted-token override and its empty-value fallback.
- Chart-name derivation against `cluster-aws`, `prometheus-meta-operator`, `chart-2-0` and long dev versions, plus the zero-archive, two-archive and unparseable cases.
- `changelog-lint` and a yamllint-equivalent pass clean; a simulated `orb pack` resolves every command reference and parameter.

Still to verify on real CI, from a consumer repo pinned to `giantswarm/architect@dev:report-chart-version-on-pr`:

- [ ] Public repo, defaults: full comment renders.
- [ ] Second push updates the same comment id, no new comment.
- [ ] Private repo: `gsociprivate` row plus the `helm registry login` caveat.
- [ ] `push_to_oci_registry: false`: OCI rows omitted, catalog row present.
- [ ] `explicit_allow_chart_name_mismatch: true`: body shows the published name.
- [ ] Tag build and post-merge default-branch build both log their skip and stay green.
- [ ] **Most important:** a repo where the App lacks the permission stays **green**. If it goes red, the default must ship as `false`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,41 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Added
+
+- `push-to-app-catalog`: new `comment_on_pr` parameter (default `true`) that reports the published
+  chart name and version on the pull request. Off-tag builds stamp
+  `X.Y.Z-dev.<branch>.<date>.<time>`, which is collision-proof but not guessable, so reviewers had to
+  open the CircleCI job and read the log to find out what to install. The comment also carries the
+  OCI reference, the digest, a `helm pull` command and a link to the git catalog.
+
+  It is a single sticky comment, found by a hidden marker keyed on the `chart` parameter and updated
+  in place on later pushes — a `PATCH` sends no notification, so a long branch produces one comment
+  instead of one per push, and a repo publishing several charts gets one comment per chart.
+
+  The pull request is resolved via `GET /repos/{o}/{r}/commits/{sha}/pulls` rather than
+  `CIRCLE_PULL_REQUEST`, which CircleCI does not provide under its GitHub App integration and which
+  is unreliable for same-repo pull requests under OAuth.
+
+  **Requires the `CircleCI architect` GitHub App to hold `Pull requests: write`** (plus
+  `Issues: write`, since pull request comments live on the issues endpoints), accepted by an
+  organisation owner. Until that is granted the step logs a warning naming the missing permission.
+  Every failure path is non-fatal by construction — no pull request yet, forked pull request (no
+  context, so no token), missing permission, rate limit — because the chart has already been
+  published by the time it runs. Nothing is posted for tag builds, default-branch builds, or when
+  both pushes are disabled.
+
+- `generate-github-token`: new `non_fatal` parameter (default `false`, so existing callers are
+  unaffected). When `true`, a missing App key or a failed `gh-token` call warns and exits 0 with
+  `GITHUB_TOKEN` unset instead of failing the step, for best-effort features that must never break a
+  build.
+
+### Fixed
+
+- `generate-github-token`: a failing `gh-token generate` left the step green with an empty token,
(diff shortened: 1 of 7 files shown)


mcharriere:
image

mproffitt (approved):

mproffitt:
I didn't see anything substantial on Claudes review so approving.

